PR #25, opened 2025-07-15
feat: add colpali retriever

paknikolai:
### Description of changes
Added Colpali retriever:

- added 3 models vidore/colSmol-256M, vidore/colpali-v1.3, vidore/colqwen2-v1.0
- added downloading models while building docker image
- added 2 more pools to process colpali model(1 for query, 1 for images) not to block other light processing embeddings, according to measurements it is possible to run query processing + image processing in parallel withought memory issues.

Added unit tests for retriever and end-to-end test:

- test for retrieving exact page from document
- test for e2e processing
- tests for requests that are running in parallel queries and images processing

Added separate docker file that allows to build image based on image of ai-dial-rag that adds saved model weights inside docker image for specified model

Some measurements for memory consumptions on T4(in colab), batch 8 seems to be best option in time/memory vise. 

Model | Type | Batch | Time (ms) | Time/Unit (ms) | Peak VRAM (GB)
-- | -- | -- | -- | -- | --
vidore/colSmol-256M | Image | 8 | 2508.89 | 313.61 | 3.99
vidore/colpali-v1.3 | Image | 8 | 2693.39 | 336.67 | 7.42
vidore/colqwen2-v1.0 | Image | 8 | 15800.93 | 1975.12 | 6.45
  |   |   |   |   |  
Model | Type | Batch | Time (ms) | Time/Unit (ms) | Peak VRAM (GB)
vidore/colSmol-256M | Image | 1 | 341.87 | 341.87 | 0.89
vidore/colSmol-256M | Image | 2 | 640.58 | 320.29 | 1.34
vidore/colSmol-256M | Image | 4 | 1310.05 | 327.51 | 2.22
vidore/colSmol-256M | Image | 8 | 2508.89 | 313.61 | 3.99

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

query + images   parallel run

Metric | colSmol-256M | colpali-v1.3 | colqwen2-v1.0
-- | -- | -- | --
Model Memory (Post-load) | 0.45 GB | 5.52 GB | 4.18 GB
Peak Memory (PyTorch) | 4.00 GB | 7.45 GB | 6.48   GB

# Model Optimization Experiments
To measure time was used vidore-benchmark, batch 4 or 8

## 0. Pytorch models
All pytorch models were run on gpu with float16

### Overview
ONNX export was attempted to optimize model inference performance by converting the PyTorch models.

### Export Results
- **vidore/colpali-v1.3** → Exportable with minor tweaks (kwargs → args).  
- **vidore/colqwen2-v1.0** → Exportable with more extensive tweaks (argument fixes, some code rewrites to fix graph tracking to enable export).  
- **vidore/colSmol-256M** → Not successful (graph not tracked, skipped to focus on colpali and colqwen models).  

### Performance (T4 & L4 GPUs)
- **Small input (query)** → 1.5–2× faster than PyTorch.  
- **Large input (image)** → ~20% slower due to memory allocations.  

### Optimizations Tried
- Replacing CPU-only nodes with GPU nodes → no big improvements, no change in allocation memory time.  
- Attention from pytorch ports as separate ops, fused into kernels → the time spent to calculate attention wasn't bottleneck and it didnt speed up mcuh the inference.  
- INT8 quantization on ONNX → no noticeable speedup, works a bit slower.  
- Different graph optimization settings → didn't result in much difference
### TensorRT
- **ONNX → TensorRT**, **ONNX(TensorRT backend)** and **Torch-TensorRT** failed (unsupported operations).  

---

## 2. Quantization (Original PyTorch Models)
- Tried libraries: **Quanto, Eetq, Hqq, TorchAo**.  
- **4-bit** → 4–10× slower (no GPU backend support on T4).  
- **INT8** → 5–10% slower than original PyTorch for all 3 models.  

---

### Backends Tested
**cudagraphs, onnxrt, openxla, tvm, inductor**
- **colpali & colsmol** → Slightly faster for image processing, but **text processing up to 10× slower**.  
- **colqwen** → moslty failed but succedded ones 2–10× slower, also Dynamic=False image inputs caused frequent recompilations.  

---


config/azure_colpali.yaml
Allob, line 1: Should it be azue_colsmal?

config/azure_colpali.yaml
```diff
@@ -0,0 +1,16 @@
+request:
+  indexing:
+    description_index: null
+    multimodal_index: null
+    colpali_index:
+      model_name: vidore/colSmol-256M
+      model_type: ColIdefics
```
Allob: image_size: 448?

aidial_rag/retrievers/colpali_retriever/colpali_index_config.py
```diff
@@ -0,0 +1,73 @@
...
+                "dimensions required by that model(ColPali: 448, ColQwen: not fixed, ColIdefics: 512)"
+            ),
+        ),
+    ]
+
+    def validate_consistency(self):
+        """validation of model name and type consistency"""
+        if self.model_name in KNOWN_MODELS:
+            expected_type = KNOWN_MODELS[self.model_name]
+            if self.model_type != expected_type:
```
Allob: If we have `KNOWN_MODELS` mapping, do we need `model_type` field in the config?
If there is only one correct `model_type` and we check it with the `KNOWN_MODELS`, it would be easier to get the correct value from the `KNOWN_MODELS` and avoid it in the config.

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,298 @@
...
+            raise RuntimeError("Processor is not initialized.")
+        batch_queries = self.processor.process_queries(queries).to(self.device)
+        with torch.no_grad():
+            with self.model_resource.get_gpu_lock():
+                query_embeddings = self.model(**batch_queries)
+        return query_embeddings
+
+    async def aembed_queries(self, queries: List[str]) -> Tensor:
+        """Async version of embed_queries that runs in CPU pool."""
+        return await run_in_indexing_cpu_pool(self.embed_queries, queries)
```
Allob: I do not fully understand how threads and locks are designed to work here.
Here the indexing_cpu_pool is used for the embed_queries. This pool has multiple threads for single-core tasks. And looks like all the threads in the pool will wait for a single gpu_lock.
For the embeddings we have a separate pools with a single process to ensure processing one-by-one - maybe we need something similar here.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,76 @@
...
+            self.set_config(config)
+
+    def get_gpu_lock(self):
+        """Get the thread lock specifically for GPU operations."""
+        return self.gpu_lock
+
+    def set_config(self, config: ColpaliIndexConfig):
+        config.validate_consistency()
+
+        with self.lock:
```
Allob: Do we need an option to modify the config for already created object?

I think, it would be better to create a new instance of the `ColpaliModelResource` if someone needs a different config for some reason, instead of modifying an existing instance. And `self.lock` will not be needed.

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,298 @@
...
+            model_resource=colpali_model_resouce,
+        )
+
+    def embed_queries(self, queries: List[str]) -> Tensor:
+        if self.processor is None:
+            raise RuntimeError("Processor is not initialized.")
+        batch_queries = self.processor.process_queries(queries).to(self.device)
+        with torch.no_grad():
+            with self.model_resource.get_gpu_lock():
+                query_embeddings = self.model(**batch_queries)
```
Allob: I suggest to move model inference into a method of a model_resource together with get_gpu_lock. It would be harder to use it incorrectly.
Probably, processor also could be moved into the model_resource .

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,298 @@
...
+    def _score_documents_with_embeddings(
+        self, query_embeddings: Tensor
+    ) -> List[Tuple[float, int]]:
+        """Score all documents against the query embeddings and return sorted (score, doc_idx) pairs."""
+        query_embeddings = query_embeddings.half()
+
+        page_scores = []
+        page_indices = []
+
+        for doc_id, doc_embedding in enumerate(self.document_embeddings):
```
Allob: This line assumes that an index of the element in the `document_embeddings` will be equal to `doc_id`. But in `from_doc_records` you are skipping the docs with `colpali_embeddings_index is None`.

I.e. for an input with a mix of the documents with image support and without image support, you can have a wrong doc_id here and could use the pages from a different document.

tests/utils/colpali_cache.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def _load_cached_data(self) -> None:
+        if not self.cache_dir:
+            return
+        query_path = self.cache_dir / f"{self.cache_key}_query_embeddings.pkl"
+        image_path = self.cache_dir / f"{self.cache_key}_image_embeddings.pkl"
+        if query_path.exists():
+            with open(query_path, "rb") as f:
+                query_data = pickle.load(f)  # noqa: S301
+                # Handle both old dict format and new list format
```
Allob: Could you, please, explain, why we have an old format here?
This cache is introduced in this very PR. It is odd that it already has some old format.

tests/utils/colpali_cache.py
```diff
@@ -0,0 +1,176 @@
...
+                pickle.dump(embeddings, f)
+
+    def __call__(self, **kwargs) -> Tensor:
+        is_image_call = "pixel_values" in kwargs
+
+        if self.use_cache:
+            # Replay mode
+            embeddings = (
+                self.recorded_image_embeddings
+                if is_image_call
```
Allob: Almost every line of the cached branch of this method has `if is_image_call` (except line 98).

Maybe the could would be easier if you just put `if is_image_call:` at the beginning. Or move logic to a function and pass embeddings and call_count parameters depending on `is_image_call`.

tests/utils/colpali_cache.py
```diff
@@ -0,0 +1,176 @@
...
+                else self.recorded_query_embeddings
+            )
+            call_count = (
+                self.image_call_count
+                if is_image_call
+                else self.query_call_count
+            )
+
+            if call_count >= len(embeddings):
+                call_count = 0
```
Allob: Why the count is set to 0 here? Why `call_count >= len(embeddings)` is not an error?

tests/utils/colpali_cache.py
```diff
@@ -0,0 +1,176 @@
...
+                if is_image_call:
+                    self.image_call_count = 0
+                else:
+                    self.query_call_count = 0
+
+            output = embeddings[call_count]
```
Allob: It is odd that the result depends on the `call_count`, not on the actual input value. If the order of the processing will change for some reason, the results from the cache will be different.

And in `ColpaliRetriever` you use `indexing_cpu_pool` which has multiple threads, which should lead to a different order of model calls in a different runs.

Allob (changes requested):

config/azure_colpali.yaml
paknikolai: changed to 512, since in config for that model it uses 512 and resized it anyway to that size, as for colpali model it uses 448

aidial_rag/retrievers/colpali_retriever/colpali_index_config.py
paknikolai: removed model_type from config, it is being mapped using KNOWN_MODELS. but i have left it to be able to easily add model of the same type but with different name

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: i have made function private, we discussed it that it would be fine if only 1 config with 1 model type will be from the beggining and it wont change. Also if allow to have multiple models at once, we will have to either have all of them in memory which would consume a lot of it(each model few gb) or unload models and load them for every request

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: outdated

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: yeah, that was an issue, added saving doc id in embedding during from_doc_records step, and then when i form chunks with metadata i use that doc_id

tests/utils/colpali_cache.py
paknikolai: it was an iteration for cache, didnt remove the comment, now  removed it

tests/utils/colpali_cache.py
paknikolai: changed it, now it is used twice to choose which embeddings to use query or image, then process is unified, and then used again to choose in which folder to save embeddings

tests/utils/colpali_cache.py
paknikolai: it was an old version of cache, it was looped. Now i use hash value from image or query, and get embeddings based on it

tests/utils/colpali_cache.py
paknikolai: yes, i have changed when i have added threads, now it uses hash value from input to get cached value

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: i have changed it, now there is a ColpaliBatchProcessor that is dealing with processing requests and forms batches. made similar to that with separate threads for embeddings and for queries. Batch is being formed by taking by one image from each pending user tasks untill batch is being formed

Dockerfile.colpali
```diff
@@ -0,0 +1,39 @@
...
+
+# Set specific model to download with default value
+ARG COLPALI_MODEL_NAME=vidore/colSmol-256M
+ENV COLPALI_MODEL_NAME=${COLPALI_MODEL_NAME}
+
+# Switch to root user for model downloads
+USER root
+
+# Copy necessary files for ColPali model download
+COPY aidial_rag/__init__.py aidial_rag/
```
Allob: Why do we need to copy aidial_rag files here? `epam/ai-dial-rag` base image should already have these files.

aidial_rag/retrievers/colpali_retriever/colpali_index_config.py
```diff
@@ -0,0 +1,20 @@
...
+class ColpaliIndexConfig(BaseModel):
+    image_size: Annotated[
+        int,
+        IndexRebuildTrigger(),
+        Field(
+            default=512,
+            description=(
+                "Specifies the size to which page images are initially resized before embedding calculation."
+                "Note: Each model's processor may further resize images to the "
+                "dimensions required by that model(ColPali: 448, ColQwen: not fixed, ColIdefics: 512)"
```
Allob: I do not fully understand this parameter. Is it really configurable (and user could change it per-request for the use case) or is it required by the the model?
If it depends on the selected model, maybe it is better to configure it together with the model?

aidial_rag/retrievers/colpali_retriever/colpali_index_config.py
```diff
@@ -0,0 +1,20 @@
...
+
+from pydantic import BaseModel, Field
+
+from aidial_rag.base_config import IndexRebuildTrigger
+
+
+class ColpaliIndexConfig(BaseModel):
+    image_size: Annotated[
+        int,
+        IndexRebuildTrigger(),
```
Allob: Does the change of this parameter makes index incompatible with the retrieval? I thought that this parameter is used during the indexing process only. I.e. the parameter is used during the indexing and affects the number of the embedding vectors we may get per page, but the list of the embeddings should be compatible with the retrieval process independent on the image_size. Or am I missing something?

config/azure_colsmol256m.yaml
```diff
@@ -0,0 +1,17 @@
...
+    chat_chain:
+      llm:
+        deployment_name: gpt-4o-2024-05-13
+        max_prompt_tokens: 16000
+    query_chain:
+      llm:
+        deployment_name: gpt-4o-2024-05-13
+        max_prompt_tokens: 8000 
+colpali_model_resource_config:
+  model_name: vidore/colSmol-256M
```
Allob: The model will be downloaded from the huggingface during the start process, isn't it?

download_model.py
```diff
@@ -1,5 +1,10 @@
 import sys
-
+import argparse
+import os
+import shutil
+from pathlib import Path
+from huggingface_hub import snapshot_download
+import tempfile
```
Allob: Is tempfile not used?

download_model.py
```diff
@@ -16,5 +21,76 @@ def download_model(name: str, path: str, *backend_args: str):
...
+    print(f"Successfully downloaded {model_name}")
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Download models for AI-Dial-RAG",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog="""
+Examples:
+  # Download a specific ColPali model
+  python download_model.py colpali /path/to/colpali_models vidore/colpali-v1.3
```
Allob: Why does the `colpali` mode use different order of arguments than `embeddings` mode?

download_model.py
```diff
@@ -16,5 +21,76 @@ def download_model(name: str, path: str, *backend_args: str):
         download_model_for_backend(name, path, backend)
 
 
+def download_colpali_model(base_path: str, model_name: str):
```
Allob: download_colpali_model looks completely different from the original download_model. Do we have any benefits by putting it in the same script?

download_model.py
```diff
@@ -16,5 +21,76 @@ def download_model(name: str, path: str, *backend_args: str):
...
+    """Download a single ColPali model from the KNOWN_MODELS map"""
+    # Import the KNOWN_MODELS map and utility functions
+    from aidial_rag.retrievers.colpali_retriever.colpali_models import (
+        KNOWN_MODELS, get_model_processor_classes, get_model_local_path, get_model_cache_path
+    )
+    
+    # Validate model name
+    if model_name not in KNOWN_MODELS:
+        print(f"Error: Model '{model_name}' not found in known models: {list(KNOWN_MODELS.keys())}")
```
Allob: Do we need a print here? The exception will be printed anyway.

download_model.py
```diff
@@ -16,5 +21,76 @@ def download_model(name: str, path: str, *backend_args: str):
...
+    cache_path = get_model_cache_path(model_path)
+    model_class, processor_class = get_model_processor_classes(model_name) 
+    model_path.mkdir(parents=True, exist_ok=True)
+    
+    # download model repository for config files and adapters weights
+    from huggingface_hub import snapshot_download
+    snapshot_download(repo_id=model_name, local_dir=model_path, local_dir_use_symlinks=False, force_download=True)
+    
+    # download base model into cache directory
+    model = model_class.from_pretrained(model_name,local_files_only=False,
```
Allob: I do not understand this code. We have both `snapshot_download` and `model_class.from_pretrained`. Which one does the download of the model?

I believe that `snapshot_download` should to the download. In this case I do not know what the `model_class.from_pretrained` is used for.

eval/eval_retriever.py
```diff
@@ -23,6 +23,9 @@
 from aidial_rag.documents import load_document_impl
 from aidial_rag.resources.dial_limited_resources import DialLimitedResources
 from aidial_rag.retrieval_chain import create_retriever
+from aidial_rag.retrievers.colpali_retriever.colpali_model_resource import (
+    ColpaliModelResource,
```
Allob: Not used?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,281 @@
...
+    model_name: Annotated[
+        str,
+        Field(
+            default="vidore/colSmol-256M",
+            description="Model name, should be one of KNOWN_MODELS keys",
+        ),
+    ]
+
+    def validate_model_name(self):
+        """Validate that model name is known"""
```
Allob, line 36: Should we use something like `model_name: Literal[KNOWN_MODELS]` or make `KNOWN_MODELS` to be  `StrEnum`?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,281 @@
...
+from aidial_rag.retrievers.colpali_retriever.colpali_models import (
+    KNOWN_MODELS,
+    get_model_cache_path,
+    get_model_local_path,
+    get_model_processor_classes,
+)
+
+# Path to pre-downloaded ColPali models for normal use in docker
+# if None model will be downloaded from Hugging Face
+COLPALI_MODELS_BASE_PATH = os.environ.get("COLPALI_MODELS_BASE_PATH", None)
```
Allob: Why are we using env variable here instead of the config?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
Allob: Do we still need a `threading.Lock` in this function?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,281 @@
...
+            for _, future in batch_items:
+                if not future.done():
+                    future.set_exception(e)
+
+
+class ColpaliModelResource:
+    def __init__(
+        self,
+        config: ColpaliModelResourceConfig | None,
+        colpali_index_config: ColpaliIndexConfig | None,
```
Allob: I do not understand, why do we need `colpali_index_config` here. I assumed that this part of the config will be used during the request processing. Could you, please, explain?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,281 @@
...
+        self.index_config: ColpaliIndexConfig | None = None
+        self.model = None
+        self.device: torch.device | None = None
+        self.processor = None
+        self.batch_processor: Optional[ColpaliBatchProcessor] = None
+        self.query_batch_processor: Optional[ColpaliBatchProcessor] = None
+        if colpali_index_config is not None and config is not None:
+            self.__set_config(config)
+
+    def __set_config(self, config: ColpaliModelResourceConfig):
```
Allob: This "constructor" for the`ColpaliModelResource` is hard to read for me.

I suggest removing the `__set_config` method and put this code in the `__init__`. And, instead of the  `__set_config` method which modifies the object state, extract several functions (with inputs and outputs). For example, the comment `# Check if local model path exists otherwise use Hugging Face` indicates that it this code could a function to handle paths, for example.

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+
+        query_embeddings_list = []
+
+        for query in queries:
+            # Add query to batch processor
+            future = await query_batch_processor.add_item(query, id)
+            query_embedding = await future
+            query_embeddings_list.append(query_embedding)
+
+        await query_batch_processor.unregister_processing_id(id)
```
Allob: `unregister_processing_id` will not be called in case of some exception. Let's use with-statement here.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,281 @@
...
+                if self.items_count < self.batch_size:
+                    should_wait = True
+                elif self.items_count == 0:
+                    break  # No more items, exit processing
+
+            # Wait to collect more items
+            if should_wait:
+                await asyncio.sleep(self.batch_wait_time)
+
+            async with self._lock:
```
Allob: Why do we need such complex logic for the batch processing?
Could you, please, explain why the existing `batched_map_with_progress` is not enough?
https://github.com/epam/ai-dial-rag/blob/development/aidial_rag/batched.py#L35

aidial_rag/retrievers/colpali_retriever/colpali_models.py
```diff
@@ -0,0 +1,75 @@
...
+    return Path(base_path) / safe_name
+
+
+def get_model_cache_path(model_path: Path) -> Path:
+    """Get the cache path for a model given model path"""
+    return model_path / "cache"
+
+
+# Path to pre-downloaded ColPali models
+COLPALI_MODELS_BASE_PATH = os.environ.get("COLPALI_MODELS_BASE_PATH", None)
```
Allob: Is it the second place where this variable is read from the env?

aidial_rag/retrievers/colpali_retriever/colpali_models.py
```diff
@@ -0,0 +1,75 @@
...
+    "vidore/colpali-v1.3": ColpaliModelType.COLPALI,
+    "vidore/colqwen2-v1.0": ColpaliModelType.COLQWEN,
+}
+
+
+def get_model_processor_classes(
+    model_name: str,
+) -> tuple[Any, Any]:
+    """Get model and processor classes for a given model name"""
+    from colpali_engine.models import (
```
Allob: colpali_engine not is not an optionale dependency. Let's move the imports to the regular place for them.

aidial_rag/retrievers/colpali_retriever/colpali_models.py
```diff
@@ -0,0 +1,75 @@
...
+# can be extended with more models if needed
+KNOWN_MODELS = {
+    # ColIdefics models
+    "vidore/colSmol-256M": ColpaliModelType.COLIDEFICS,
+    "vidore/colpali-v1.3": ColpaliModelType.COLPALI,
+    "vidore/colqwen2-v1.0": ColpaliModelType.COLQWEN,
+}
+
+
+def get_model_processor_classes(
```
Allob, line 46: Do we need to expose the function that will return model and processor classes? We can have a function which will return loaded model and processor instead.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,281 @@
...
+            if COLPALI_MODELS_BASE_PATH:
+                local_model_path = get_model_local_path(
+                    COLPALI_MODELS_BASE_PATH, model_name
+                )
+                if os.path.exists(local_model_path):
+                    model_name = local_model_path
+                    cache_path = get_model_cache_path(local_model_path)
+                    print(f"loading model from local path: {local_model_path}")
+                else:
+                    print("loading model from hugging face")
```
Allob: If the local path is specified and the model is expected to be loaded from local path, I think, we should throw an error here instead of a fallback to the huggingface.

aidial_rag/retrievers/colpali_retriever/colpali_models.py
```diff
@@ -0,0 +1,75 @@
...
+
+def get_model_local_path(base_path: str, model_name: str) -> Path:
+    """Get the local path for a model given base path and model name"""
+    safe_name = get_safe_model_name(model_name)
+    return Path(base_path) / safe_name
+
+
+def get_model_cache_path(model_path: Path) -> Path:
+    """Get the cache path for a model given model path"""
+    return model_path / "cache"
```
Allob, line 74: You use a `model_local_path` as a `model_path` here for a cache. Wouldn't the cache folder for the huggingface also contain the the model path inside?

aidial_rag/retrievers/colpali_retriever/colpali_models.py
```diff
@@ -0,0 +1,75 @@
...
+            return ColIdefics3, ColIdefics3Processor
+        case ColpaliModelType.COLQWEN:
+            return ColQwen2, ColQwen2Processor
+        case _:
+            raise ValueError("Invalid ColPali model type")
+
+
+def get_safe_model_name(model_name: str) -> str:
+    """Convert model name to safe directory name"""
+    return model_name.replace("/", "_")
```
Allob: Why do we need this? It looks odd if we a trying to construct the model path during the load.

I would prefer one of the two following approaches:
1. We just specify ready-to-use path to the model in the config (like for embeddings) - no path transformations needed, less code, less places for errors.
2. We just specify a single cache folder for the huggingface library, and expect it to handle the paths by the model names. Again, no path transformations is needed from our side.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: it is possible to adapt it, but then we wont have control over batch, since batch there forms only from 1 user, but profit from batching isnt that big so i guess we wont lose that much

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+        futures = []
+        async for image in images.agen:
+            future = await batch_processor.add_item(
+                image, id
+            )  # await to get the Future
+            futures.append(future)
+
+        # Wait for all futures to complete
+        with TqdmProgressBar(total=len(futures), file=stageio) as pbar:
+            for future in futures:
```
Allob: You can use TqdmProgressBar as a reqular tqdm:
`for future in TqdmProgressBar(futures, file=stageio):`
And no need for manual call for `pbar.update()`.

aidial_rag/retrievers/colpali_retriever/colpali_models.py
paknikolai: i guess yes we can use the model name too, it will create additional folder, because names for models are like vidore/colSmol-256M, origianlly i wanted to make it to be one name

config/azure_colsmol256m.yaml
paknikolai: if it wasnt downloaded before then yes. it checks the env variable which is being set in docker file, if it wasnt set, then it downloads from hugging face, if it was set then it loads it from path where model was saved during building step

download_model.py
paknikolai: it doesnt benefit from it just not to create different file, because it is also predownloading model weights, if needed i can create separate file

download_model.py
paknikolai: it is complicated download since model repo doesnt contain base model weights. so snapshot would only download   configs and weights for adapters. from pretrained needed to trigger downloading base model weights(llm and visual encoders) into cache folder

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+        image_embeddings_list = []
+
+        stageio.write("Processing images\n")
+
+        id = await batch_processor.register_processing_id()
+
+        # Collect all futures first
+        futures = []
+        async for image in images.agen:
+            future = await batch_processor.add_item(
```
Allob: Here we will unwind the whole `images.agen` and will have the images.
The point of the async generator was to avoid keeping all the page images in memory at the same time. For pdfs with 500-1000 pages this could be significant.

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+        # Wait for all futures to complete
+        with TqdmProgressBar(total=len(futures), file=stageio) as pbar:
+            for future in futures:
+                image_embedding = await future
+                image_embeddings_list.append(image_embedding)
+                pbar.update()
+
+        await batch_processor.unregister_processing_id(id)
+
+        # Pad embeddings to same shape
```
Allob: Could you, please, explain, why do we need this?
We know that all the source images will have the same size `colpali_index_config.image_size`.
Why would  the embeddings have the different shape?

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+                "ColpaliModelResource did not return a valid device."
+            )
+        if model is None:
+            raise RuntimeError(
+                "ColpaliModelResource did not return a valid model."
+            )
+
+        # Get or create batch processor with processing method
+        batch_processor = colpali_model_resource.get_batch_processor(
+            lambda images: ColpaliRetriever._process_images_batch(
```
Allob: Here we get `model, processor, device` from `colpali_model_resource`, also get `get_batch_processor` and do
```
        with torch.no_grad():
            outputs = model(**inputs)
```

I think all of this could be hidden into colpali_model_resource or it's batch_processor.

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+                images, processor, model, device
+            ),
+            pool_func=run_in_heavy_indexing_embeddings_pool,
+        )
+
+        image_embeddings_list = []
+
+        stageio.write("Processing images\n")
+
+        id = await batch_processor.register_processing_id()
```
Allob: Instead of registering/unregistering and the passing the correct id to batch_processor, we could have here an instance of a batch_processor which would already have an id inside. The id could be registered by the call get_batch_processor.
```
with colpali_model_resource.get_batch_processor() as batch_processor:
    ....
    async for image in images.agen:
        future = await batch_processor.add_item(image)
    ...
```

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
```diff
@@ -0,0 +1,372 @@
...
+    def _process_images_batch(
+        images_batch: List[str], processor, model, device
+    ) -> List[torch.Tensor]:
+        """Process a batch of images using the ColPali model on GPU."""
+        # Convert base64 images to PIL images
+
+        # process images with processor
+        pil_images = []
+        for image in images_batch:
+            pil_image = pil_image_from_base64(image)
```
Allob, line 222: I think we need a version of the `extract_page_images()` which will return pil.Image instead of base64 string.
Until now, both description retriever and multimodal retriever use images as base64 strings.
See https://github.com/epam/ai-dial-rag/blob/db5e14bebcc41600c83b208c065bdbe3be8d1cdb/aidial_rag/retrievers/page_image_retriever_utils.py#L43

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: it was done the same way as for embeddings, there they also use env variable in docker to donwload models to folder  and then use that folder from variable in code to load it later

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: it is not needed actually, since all the parameters now in ColpaliModelResourceConfig, now ColpaliIndexConfig is used only as indicator to build index or not(the same way as other indexes do, if not none then build)

Dockerfile.colpali
paknikolai: indeed here it has already been copied in base image, left only copying downloading script

aidial_rag/retrievers/colpali_retriever/colpali_index_config.py
paknikolai: removed this parameter since it is optional for laoding extract_page_images and only used there

aidial_rag/retrievers/colpali_retriever/colpali_index_config.py
paknikolai: it was used only for extract_page_images, but there this parameter is optional so i have removed it

download_model.py
paknikolai: removed

download_model.py
paknikolai: changed to match embeddings

download_model.py
paknikolai: removed, raise error would be enough

eval/eval_retriever.py
paknikolai: removed

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: made separate str enum for model names KnownModels, and renamed mapping to  MODEL_NAME_TO_TYPE

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: removed set config, all is done in init function. made separate function load_model_and_processor to load model and processors by model name

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: not relevant sinc, now utilizing simplier approach

aidial_rag/retrievers/colpali_retriever/colpali_models.py
paknikolai: removed

aidial_rag/retrievers/colpali_retriever/colpali_models.py
paknikolai: moved to header

aidial_rag/retrievers/colpali_retriever/colpali_models.py
paknikolai: added such function, now it loads it by model name

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: changed behaviour, now it throws an error

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: here yes, but not it is not relevant, since approach was simplified

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: i guess it was done for colqwen, since it can have different image sizes unlice colpali and colsol

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: hid it in coplali model resource, now direct use of processor and model is not needed, since added some wrapping function to call them

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: not relevant since we have changed the approach

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: do you suggest to change or add function to retunr images instead of base64 string?

aidial_rag/retrievers/colpali_retriever/colpali_models.py
paknikolai: no, it uses default cache dir so it should be set explicitly

download_model.py
Allob: If `model_class.from_pretrained` will download the model including the base model, then maybe `snapshot_download`  is not needed?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
Allob: Embeddings is an old code which hasn't been rewritten to use the config yet.
For the new code, let's put this parameter to the AppConfig. ColpaliModelResource is already created in the app, so it should be easy to do correctly.

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
Allob: What do you mean by `different image sizes` ? Do you mean that different pages from the same document are expected to have different sizes?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,131 @@
...
+    def calculate_queries_embeddings(self, queries: List[str]) -> List[Tensor]:
+        """Embed queries using the ColPali model."""
+        self._check_model_processor_device_is_set()
+        assert self.processor is not None
+        # Process queries with ColPali
+        inputs = self.processor.process_queries(queries).to(self.device)
+
+        return self._run_model(inputs)
+
+    def calculate_images_embeddings(self, images: List[str]) -> List[Tensor]:
```
Allob: `images: List[str]` - is it actually str or pil image?

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
Allob: I still do not understand why the `colpali_index_config` has to be passed here (in the constructor of the ColpaliModelResource instance). Especially if it is not actually needed, as you said.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+        self.model = None
+        self.device: torch.device | None = None
+        self.processor = None
+        # if both are set then we can load model
+        if (
+            colpali_index_config is not None
+            and colpali_index_config.enabled
+            and model_resource_config is not None
+        ):
```
Allob: If one of these fields is not set, we will get a created instance of a ColpaliModelResource, which is not functional.

And we have to call `_check_model_processor_device_is_set` every time when we need to use it, because it can be in 2 states: valid one and an invalid one.

The code will be easier to read and use if the ColpaliModelResourceConfig could have only valid state. I.e. if ColpaliModelResourceConfig is not set, we can just not create an instance of ColpaliModelResourceConfig at all (this can be done in app.py). And if the ColpaliModelResourceConfig, we should assume that it is in a valid state to use if needed (otherwise it should throw an exception from the constructor if something goes wrong).

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
```diff
@@ -0,0 +1,131 @@
...
+        self.model_resource_config: ColpaliModelResourceConfig | None = (
+            model_resource_config
+        )
+        self.model = None
+        self.device: torch.device | None = None
+        self.processor = None
+        # if both are set then we can load model
+        if (
+            colpali_index_config is not None
+            and colpali_index_config.enabled
```
Allob: `colpali_index_config` and `colpali_index_config.enabled` values come from the part of the config which could be overridden on per-request basis: https://github.com/epam/ai-dial-rag/pull/25/files#diff-50338faca97bed0430a4330eb9445199dd51a3c636bf95bee711680786b234f4R211

I.e. if the default value for the colpali_index_config.enabled is False, and someone will enable in a particular request, the request will fail, because the ColpaliModelResourceConfig will not be in a valid usable state.

download_model.py
Allob: Before this change the download_model.py was a simple script and it was easy to understand how the embeddings model was downloaded. Now it is harder to understand it, and there is no benefit of having some shared parts.
If would be much easier to have a separate download script instead of commands and subparsers.

tests/test_colpali_retriever.py
```diff
@@ -0,0 +1,455 @@
...
+@pytest.mark.asyncio
+@e2e_test(filenames=("alps_wiki.pdf",))
+async def test_colpali_retrieval_e2e(attachments):
+    """
+    End-to-end test for ColPali retrieval through the full app.
+    Tests the complete flow from document upload to answer generation.
+    """
+    # Patch create_retriever to use only ColPali retriever
+    with patch(
+        "aidial_rag.retrieval_chain.create_retriever", new=mock_create_retriever
```
Allob, line 248: I do not understand why we need to mock `create_retriever` here. The run_e2e_test function already replaces the ColpaliModelResource with CachedColpaliModelResource. Shouldn't it be enough to get the responses from cache?

tests/test_colpali_retriever.py
```diff
@@ -0,0 +1,455 @@
...
+    from fastapi.testclient import TestClient
+
+    from aidial_rag.app import create_app
+
+    # Create app with cached model resource
+    app_config = create_colpali_only_config()
+    cached_app_class = create_cached_app_config()
+
+    # Patch the app creation to use cached model resource
+    with patch("aidial_rag.app.DialRAGApplication", new=cached_app_class):
```
Allob: We can just add a parameter to the DialRAGApplication, which would allow to inject the dependency on the ColpaliModelResource which should be used here, instead of patching.

tests/cache/test_colpali_retriever/ec034ed955b0303d2dbace1b0ea8ba5e_query_embeddings_cache.pkl
Allob, line 1: Looks like you forgot to include the name of the file with the tests into the path. This may cause collisions for the tests with the same name in a different files.
See `tests/cache/test_colpali_retriever/test_colpali_retrieval_e2e/1adc73f3b7d87d911383d346295d6d66.response`.

paknikolai: moved it to a separate folder, but this embeddings are for one pdf which i use in tests and it is shared so cant use test name in it.

tests/test_colpali_retriever.py
paknikolai: removed cached app class.

adding class which is used doesnt seem that good decision, and adding resource as a parameter to app also seems to be odd.
also i have tried replace the resource instance afrer creation of DialRAGApplication but getting instance of it from app is not that simple, so i instead just patched class that was used when colpali resource is creating, so instead standard one it  is using cahced version

tests/test_colpali_retriever.py
paknikolai: i have simplified mock_create_retriever, colpali model resource is being used from arguments rn, which should be CachedColpaliModelResource which was set in the app.

there is a problem with create_retriever, if pdf is small then it will be using AllDocumentsRetriever instead, and if pdf is big enough then it will add semantic retriever anyway, so here i explicitly return only colplali retriever.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: indeed, now i have removed dependency on it from colapli_model_resource, now it depends only on ColpaliModelResourceConfig, and now it creates model all the time and its field cannot be None anymore, but now ColpaliModelResource itself can be None

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: removed invalid state and simplified logic, now it ColpaliModelResource itself can be either none or valid depending on ColpaliModelResourceConfig which is set using app config on start.

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: yes list of pil images, corrected that

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: now i have removed it, before that was a logic where when index config was passed and if it was none then resource wasnt allocated, which turned to be invalid logic, now ColpaliModelResource is independent from index config and depends only on ColpaliModelResourceConfig

aidial_rag/retrievers/colpali_retriever/colpali_model_resource.py
paknikolai: all locks have been removed since allocation of model is being done on start of the app

download_model.py
paknikolai: no, since from pretrained  doesnt save anything into desired folder. to save save_from_pretrained should be called but it saves only model related weight and configs, but misses processors related configs, so better to donwload the whole repo  not to miss needed files

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: not relevant, now it froms batch then wait for it to complete and then continue to process

aidial_rag/retrievers/colpali_retriever/colpali_retriever.py
paknikolai: pdf  can have different sizes, and in case colqwen model it will produce embedding with different length

paknikolai: removed padding since they are not needed

aidial_rag/retrievers/colpali_retriever/colpali_models.py
```diff
@@ -0,0 +1,108 @@
...
+        model_name,
+        torch_dtype=torch.float16,
+        device_map=device,
+        cache_dir=cache_path
+        if cache_path
+        else None,  # cache containt base models weights
+        local_files_only=model_path
+        is not None,  # if set use only local files from folder
+    ).eval()
+    processor = processor_class.from_pretrained(model_name)
```
Allob, line 88: Why is the `cache_dir` not needed for `processor_class.from_pretrained`?

paknikolai: here cache dir has base weights for image feature extractor and llm, and all processor's configs are in the model folder which was donwloaded using snapshot download, so cache in this case is not needed

download_model.py
Allob: I still do not understand what was the idea here and how it works.
You said `so snapshot would only download configs and weights for adapters. from pretrained needed to trigger downloading base model weights(llm and visual encoders) into cache folder`.
Also you said `from pretrained doesnt save anything into desired folder`.

The code in the [colpali_models.py](https://github.com/epam/ai-dial-rag/pull/25/files#diff-5e6217c41fef1e3e9134f391c30512dfd4886d08c8dba62b26b7ad04ebad0b26) could be used to download the model and the processor using `from_pretrained` and without using the `snapshot_download`. So, it could be possible to use the same code here.

If I understood correctly, you are mixing two different approaches here: download model to the cache (and put the cache to the docker image) and download model with `snapshot_download` specifying the local_path explicitly. Also, if I understood correctly, we have 3 components here: model, base model and processor. But I do not understand, why we are not using the same approach for all 3 components here. I believe, we should be able to explicitly make `snapshot_download` both for model or base model, without using `from_pretrained`; or, alternatively, use a single cache dir for model, base model and processor, without using `snapshot_download`. In both cases it will be easier to understand that the mixed approach.

paknikolai: donwloading such complex model is also complex process, just using save_pretrained doesnt work since it saves only wrappers configs and its weights, but not the base models that are used. technically all base models also can be saved in such way, but then we would have manually for each layer of models change all configs(which has path to base model or its name to the one we have saved) to be able to load it using from_pretrained(otherwise it will try to load base model weight from huggingface) and also for each model we should manually parse configs and see which base model it uses to save it, i find such approach too complex

So instead i save all configs which are needed for the wrapper(it also includes processors configs) using snapshot download and then when i set cache_folder to from_pretrained method, it saves all the rest needed in that cache and i dont need to modify any configs to load this weights.

code from loading the model cannot be reused, it has different flags so i think it is better to keep them separate not to overcomplicate them with additional parameters for loading and saving

Allob: Ok, you can keep the code separate because of the flags. But still, why not save everything to the cache_folder using from_pretrained ?

paknikolai: from pretrain doesnt save base model and there is no argument to do so, and only some configs that are related to model itself are saved, but other configs that are not model related won't be saved, for example configs for processor will be lost and model after loading wont properly work

Allob: I'm even more confused. Before you said that `so snapshot would only download configs and weights for adapters. from pretrained needed to trigger downloading base model weights`, but now you are saying `from pretrain doesnt save base model`.

I've checked it myself using the following script:
```python
from colpali_engine.models import (
    ColIdefics3,
    ColIdefics3Processor,
)

model_name = "vidore/colSmol-256M"

сache_dir = "./models/colsmol"

processor = ColIdefics3Processor.from_pretrained(model_name, cache_dir=сache_dir)
model = ColIdefics3.from_pretrained(model_name, cache_dir=сache_dir)
```

And after running, I got the following files in the `./models/colsmol`:
```
$ find ./models/colsmol/
./models/colsmol/
./models/colsmol/.locks
./models/colsmol/.locks/models--vidore--colSmol-256M
./models/colsmol/.locks/models--vidore--ColSmolVLM-Instruct-256M-base
./models/colsmol/models--vidore--colSmol-256M
./models/colsmol/models--vidore--colSmol-256M/.no_exist
./models/colsmol/models--vidore--colSmol-256M/.no_exist/a59110fdf114638b8018e6c9a018907e12f14855
./models/colsmol/models--vidore--colSmol-256M/.no_exist/a59110fdf114638b8018e6c9a018907e12f14855/audio_tokenizer_config.json
./models/colsmol/models--vidore--colSmol-256M/.no_exist/a59110fdf114638b8018e6c9a018907e12f14855/chat_template.jinja
./models/colsmol/models--vidore--colSmol-256M/blobs
./models/colsmol/models--vidore--colSmol-256M/refs
./models/colsmol/models--vidore--colSmol-256M/refs/main
./models/colsmol/models--vidore--colSmol-256M/snapshots
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/adapter_config.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/added_tokens.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/chat_template.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/config.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/merges.txt
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/preprocessor_config.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/processor_config.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/special_tokens_map.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/tokenizer.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/tokenizer_config.json
./models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855/vocab.json
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/blobs
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/refs
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/refs/main
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/snapshots
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/snapshots/99ca96f1f6b95b3a69e6abef74a2416cb738fed0
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/snapshots/99ca96f1f6b95b3a69e6abef74a2416cb738fed0/config.json
./models/colsmol/models--vidore--ColSmolVLM-Instruct-256M-base/snapshots/99ca96f1f6b95b3a69e6abef74a2416cb738fed0/model.safetensors
```
It looks like the everything was downloaded: base model ColSmolVLM-Instruct-256M and colSmol-256M, including the processor_config.json.

paknikolai: i guess you are right about saving some of the base model configs, but  i have tried to load model from that snapshot folder and it lacks adapters weights(tried colsol and colqwen)

```python
# different directory for snapshot
dir = './models/colsmol/models--vidore--colSmol-256M/snapshots/a59110fdf114638b8018e6c9a018907e12f14855'

processor = ColIdefics3Processor.from_pretrained(dir, cache_dir=сache_dir, local_files_only=True)
model = ColIdefics3.from_pretrained(dir, cache_dir=сache_dir, local_files_only=True)`
```
i get error that 
```text
File "/home/nikolai/code/ai-dial-rag/.venv/lib/python3.11/site-packages/transformers/modeling_utils.py", line 4947, in from_pretrained
    model.load_adapter(
  File "/home/nikolai/code/ai-dial-rag/.venv/lib/python3.11/site-packages/transformers/integrations/peft.py", line 250, in load_adapter
    adapter_state_dict = load_peft_weights(peft_model_id, token=token, device=device, **adapter_kwargs)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/nikolai/code/ai-dial-rag/.venv/lib/python3.11/site-packages/peft/utils/save_and_load.py", line 554, in load_peft_weights
    has_remote_safetensors_file = file_exists(
...
    ```
    
  So to make it work i need to save also explicitly adapters weights there + some code to locate the snapshot folder with all configs, i think just using snapshot_download is much easier

Comment (Allob) [reply download_model.py]: Try the same code as for the download model:
```
model_name = "vidore/colSmol-256M"

сache_dir = "./models/colsmol"

processor = ColIdefics3Processor.from_pretrained(model_name, cache_dir=сache_dir)
model = ColIdefics3.from_pretrained(model_name, cache_dir=сache_dir)
```
If you run this code after the model already have been downloaded to the сache_dir, it will be loaded without download. You can add the `local_files_only=True` parameter to make sure that no download will be made.

I.e. you have changed the "model name" - you got some issues with matching the new "model name" to the paths for the existing files in the cache.

I've already asked the questions in the compali_models.py:
https://github.com/epam/ai-dial-rag/pull/25/files/becad27e7c47584f8310cf3969fcd8e4b5b844c2#r2334039569
https://github.com/epam/ai-dial-rag/pull/25/files/becad27e7c47584f8310cf3969fcd8e4b5b844c2#r2334024863
From my point of view, by trying to interfere with the model cache folders manually, you have created more difficulties you have to solve.

Comment (paknikolai) [reply download_model.py]: I am using the cache since all other attempts to save model didnt succeed, since some were way too difficult and complex, or other just didnt save everything needed to be saved to be able to load it without access to huggingface.


using exact same code as you have provided, with flag local_files_only=True and no internet connection, i get this error 
```text
requests.exceptions.ConnectionError: (MaxRetryError('HTTPSConnectionPool(host=\'huggingface.co\', port=443): Max retries exceeded with url: /vidore/colSmol-256M/resolve/main/adapter_model.safetensors (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x75f2d5bc40d0>: Failed to resolve \'huggingface.co\' ([Errno -3] Temporary failure in name resolution)"))'), '(Request ID: c54abe2e-3e24-407d-85ba-4a9798e85f43)')
```

with internet connection this code works fine.

but if i use safe_pretrained or use snapshot download and put that folder instead model name + cahce then it load model without internet connection too

Allob: According to the https://huggingface.co/docs/transformers/installation#offline-mode , this could be solved by using `HF_HUB_OFFLINE=1`.
Just checked it with no internet connection - worked for me.

Also, you are using `snapshot_download` conly to get an adapter model. I.e. you are using `model_class.from_pretrained` to download the base model, and I would expect the same `ConnectionError` issue there without the `HF_HUB_OFFLINE=1`.

paknikolai: with HF_HUB_OFFLINE=1 it also works for me, but then setting this variable in the app would be tricky since it should be set before all hugging face imports, so it will be one for the entire app not just for this part of code.
Also setting this variable depends on app config(has path for model or not) which would be hard to make since all hugging face import should be done after reading that config

paknikolai: simplified it, now i use HF_HOME to make huggingface to save cache to specific folder, then i load model and it saves cache, then i transfer cache to image, also had to set HF_HUB_OFFLINE=1 to force huggingface to look into cache otherweise it was trying to get data from repo which casued error if it is not available. loading is simple, just calling from_pretrained method

PR #27, opened 2025-07-15
fix: extract retrieval to a separate chain

Allob:
### Description of changes

- move retrieval to a separate chain to use separately in the future
- introduce RetrievalResults class which passes the results between retrieval and generation and will be used in the Retrieval Api in the future

### Checklist

### DIFF
--- a/aidial_rag/app.py
+++ b/aidial_rag/app.py
@@ -1,13 +1,20 @@
 import asyncio
 import logging
 from contextlib import asynccontextmanager
-from typing import List, Tuple
+from typing import Any, Dict, List, Tuple
 
 from aidial_sdk import DIALApp, HTTPException
-from aidial_sdk.chat_completion import ChatCompletion, Choice, Request, Response
+from aidial_sdk.chat_completion import (
+    ChatCompletion,
+    Choice,
+    Message,
+    Request,
+    Response,
+)
 from langchain.retrievers import EnsembleRetriever
 from langchain.schema import BaseRetriever, Document
 from langchain_core.retrievers import RetrieverLike
+from langchain_core.runnables import Runnable
 from pydantic import ValidationError
 
 from aidial_rag.app_config import AppConfig, RequestConfig
@@ -34,7 +41,7 @@ from aidial_rag.repository_digest import (
     RepositoryDigest,
     read_repository_digest,
 )
-from aidial_rag.request_context import create_request_context
+from aidial_rag.request_context import RequestContext, create_request_context
 from aidial_rag.resources.cpu_pools import init_cpu_pools
 from aidial_rag.retrieval_chain import create_retrieval_chain
 from aidial_rag.retrievers.all_documents_retriever import AllDocumentsRetriever
@@ -212,11 +219,21 @@ def get_configuration(request: Request) -> dict:
 
 
 async def _run_rag(
-    request_context, choice, request_config, retrieval_chain, chain_input
+    request_context: RequestContext,
+    request_config: RequestConfig,
+    retrieval_chain: Runnable[Dict[str, Any], Dict[str, Any]],
+    messages: List[Message],
+    document_records: List[DocumentRecord],
 ):
+    choice = request_context.choice
+    chain_input = {
+        "chat_history": transform_history(messages),
+        "chat_chain_config": request_config.qa_chain.chat_chain,
+        "doc_records": document_records,
+    }
+
     reference_items = await generate_answer(
         request_context=request_context,
-        request_config=request_config,
         retrieval_chain=retrieval_chain,
         chain_input=chain_input,
         content_callback=choice.append_content,
@@ -365,18 +382,12 @@ class DialRAGApplication(ChatCompletion):
                 )
 
             with profiler_if_enabled(choice, request_config.use_profiler):
-                chain_input = {
-                    "chat_history": transform_history(messages),
-                    "chat_chain_config": request_config.qa_chain.chat_chain,
-                    "doc_records": document_records,
-                }
-
                 await _run_rag(
                     request_context,
-                    choice,
                     request_config,
                     retrieval_chain,
-                    chain_input,
+                    messages,
+                    document_records,
                 )
 
     async def configuration(self, request):
--- a/aidial_rag/app.py
+++ b/aidial_rag/app.py
@@ -29,12 +29,14 @@ from aidial_rag.documents import load_documents
 from aidial_rag.index_record import ChunkMetadata, RetrievalType
 from aidial_rag.index_storage import IndexStorage
 from aidial_rag.qa_chain import generate_answer
+from aidial_rag.query_chain import create_get_query_chain
 from aidial_rag.repository_digest import (
     RepositoryDigest,
     read_repository_digest,
 )
 from aidial_rag.request_context import create_request_context
 from aidial_rag.resources.cpu_pools import init_cpu_pools
+from aidial_rag.retrieval_chain import create_retrieval_chain
 from aidial_rag.retrievers.all_documents_retriever import AllDocumentsRetriever
 from aidial_rag.retrievers.bm25_retriever import BM25Retriever
 from aidial_rag.retrievers.description_retriever.description_retriever import (
@@ -46,6 +48,7 @@ from aidial_rag.retrievers.multimodal_retriever import (
 )
 from aidial_rag.retrievers.semantic_retriever import SemanticRetriever
 from aidial_rag.stages import RetrieverStage
+from aidial_rag.transform_history import transform_history
 from aidial_rag.utils import profiler_if_enabled, timed_stage
 
 APP_NAME = "dial-rag"
@@ -208,6 +211,26 @@ def get_configuration(request: Request) -> dict:
     return custom_configuration_dict
 
 
+async def _run_rag(
+    request_context, choice, request_config, retrieval_chain, chain_input
+):
+    reference_items = await generate_answer(
+        request_context=request_context,
+        request_config=request_config,
+        retrieval_chain=retrieval_chain,
+        chain_input=chain_input,
+        content_callback=choice.append_content,
+    )
+
+    document_records = chain_input["doc_records"]
+    # Answer has already been streamed to the user, so we don't need to do anything here.
+    for i, reference_item in enumerate(reference_items):
+        if attachment := doc_to_attach(
+            reference_item, document_records, index=(i + 1)
+        ):
+            choice.add_attachment(**attachment)
+
+
 class DialRAGApplication(ChatCompletion):
     app_config: AppConfig
     index_storage: IndexStorage
@@ -310,16 +333,6 @@ class DialRAGApplication(ChatCompletion):
                 )
                 return
 
-            with timed_stage(choice, "Prepare indexes for search"):
-                retriever = await loop.run_in_executor(
-                    None,
-                    create_retriever,
-                    request_context.choice,
-                    request_context.dial_config,
-                    document_records,
-                    request_config.indexing.multimodal_index,
-                )
-
             last_message_content = messages[-1].content
             if last_message_content is None:
                 return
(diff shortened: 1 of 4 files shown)


aidial_rag/qa_chain.py
```diff
@@ -210,36 +205,20 @@ async def get_reference_documents(chain_input, chain) -> AsyncIterator:  # noqa:
 
 async def generate_answer(
     request_context: RequestContext,
-    qa_chain_config: QAChainConfig,
-    retriever: BaseRetriever,
-    messages: List[Message],
+    request_config: RequestConfig,
```
adubovik: Why passing the whole `RequestConfig` instead of the part of it that `generate_answer` is actually needed: `RequestConfig.qa_chain`?

aidial_rag/app.py
```diff
@@ -208,6 +211,26 @@ def get_configuration(request: Request) -> dict:
     return custom_configuration_dict
 
 
+async def _run_rag(
+    request_context, choice, request_config, retrieval_chain, chain_input
```
adubovik: Please add type annotations for the method arguments.

aidial_rag/app.py
```diff
@@ -320,33 +355,29 @@ async def chat_completion(
...
+                    "doc_records": document_records,
+                }
+
+                await _run_rag(
+                    request_context,
+                    choice,
+                    request_config,
+                    retrieval_chain,
+                    chain_input,
+                )
```
adubovik: The data is denormalized - the same config is passed to the `_run_rag` method twice, which is confusing:

```python
chain_input["chat_chain_config"] == request_config.qa_chain.chat_chain
```

Why not augment `chain_input` at the very last moment upstream? When it's really necessary _(that is when passing the whole `request_config` becomes not necessary)_.

aidial_rag/retrieval_chain.py
```diff
@@ -96,4 +98,65 @@ async def create_image_by_page(
...
 
-# TODO: Implement retrieval chain here
+@chain
+async def create_retrieval_results(
+    input: dict,
+) -> RetrievalResults:
+    """Create retrieval results from the input data."""
+    doc_records: List[DocumentRecord] = input.get("doc_records", [])
+    index_items: List[Document] = input.get("found_items", [])
+    image_by_page: dict = input.get("image_by_page", {})
```
adubovik: `image_by_page: Dict[Tuple[int, int|None], str]`

aidial_rag/retrieval_chain.py
```diff
@@ -96,4 +98,65 @@ async def create_image_by_page(
     return image_by_page
 
 
-# TODO: Implement retrieval chain here
+@chain
+async def create_retrieval_results(
+    input: dict,
```
adubovik: Nit: untyped dicts make me sad 😢

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,63 @@
...
+
+    CONTENT_TYPE: ClassVar[str] = (
+        "application/vnd.aidial.rag.retrieval-results+json"
+    )
+
+    class Chunk(BaseModel):
+        """Chunk of the document retrieved by the retriever."""
+
+        doc_id: int = Field(
+            description="Index of the attached document in the request, starting from 0."
```
adubovik: 1. Opt: `starting from 0.` => `0-based index`, reads better imo
2. Is it an index of the attachment in the whole request, not only the last message, right?

```
User: bla-bla; attachments: [doc_id=0,doc_id=1,doc_id=2]
Bot : bla-bla; attachments: [doc_id=3,doc_id=4]
User: bla-bla; attachments: [doc_id=5]
```

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+        chunk_id: int = Field(
+            description="Index of the chunk in the document, starting from 0."
+        )
+        source: str = Field(
```
adubovik, line 22: If it's always an URL, maybe call the field as such? `url`, `source_url`, `doc_url`.

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,63 @@
...
+            description="URL to the document attachment (relative URL to the Dial storage, or full url for an external document), could have an url fragments like #page=3."
+        )
+        source_display_name: str | None = Field(
```
adubovik, line 28: `doc_display_name`? in line with `doc_id`

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,63 @@
...
+            default=None,
+            description="Human-readable name of the document (the same that Dial RAG displays in the stages).",
+        )
+        text: str | None = Field(
+            default=None,
+            description="Text of the chunk, may be empty, for example, for an image.",
+        )
+        page_number: int | None = Field(
```
adubovik, line 36: `page_index`? 
`page_number` could be confused with the number of pages.

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,63 @@
...
+            default=None,
+            description="The number of the page in the document, the chunk belongs to, if applicable. "
```
adubovik: `the number of X` usually refers to a quantity.
`page number` or `page index` is less ambiguous.

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,63 @@
...
+            "Starting from 1, i.e. the same as in the page fragment of the URL.",
+        )
+        page_image: int | None = Field(
```
adubovik: 1. Maybe `image_index`? `page_image` implies that the field contains the image itself.


2. Is it correct that either text is not None or page_image is not None?

If so, then we can introduce a content field that will hold **either** text **or** a reference to an image:

```python
content: ChunkTextContent | ChunkImageContent
```

If not, we could at least highlight that these fields denote the **content** of the chunk:

```python
content_text: str | None
content_image_index: int | None
```

3. Why not point to an image itself rather than having an indirection? Each image belongs to one and only one chunk, right?

aidial_rag/retrieval_chain.py
```diff
@@ -96,4 +98,65 @@ async def create_image_by_page(
...
+            source=chunk.metadata["source"],
+            source_display_name=chunk.metadata.get("source_display_name"),
+            page_number=chunk.metadata.get("page_number"),
+        )
+
+        image_key = (
+            doc_id,
+            chunk.metadata.get("page_number"),
+        )
+        if image_key in image_by_page and image_key not in used_image_keys:
```
adubovik: So if the image spans multiple chunks, then it will be attached to the very first one _(with the smallest page number presumably)_. The rest of the chunks won't reference the image they overlap with, right?

aidial_rag/retrieval_api.py
Allob: You have a nice point here.
For now doc_id is an in doc_records array and attachment_links array collected from the history at the beginning of the request. It works for this PR, because the RetrievalResults is used internally during the processing request.

You are correct that we could have an issue with the order of the attachments here. For example, the order between the attachments from the custom_content and from the `/attrach` command is not specified here.

I suggest leaving doc_id for this PR, and change it later. I'll discuss with Uladzislau and, probably, change the index to the url from the attachment_link later.

aidial_rag/retrieval_api.py
Allob: > 1. Maybe image_index? page_image implies that the field contains the image itself.

`page_image` is used instead of the `image_index`, because I assume that the images could be used not only as an image of the whole page. Fox example, for html or markdown we may want to to pass the images embedded in the text.

> 2. Is it correct that either text is not None or page_image is not None?

Both `text` and `page_image` could be not None.  `content_image_index` assumes that image includes this chunk only, but it current RAG we have the image of the whole page.

> 3. Each image belongs to one and only one chunk, right?

We may have multiple chunks for a single page. I.e. multiple chunks could share the same page image, and I want it to be supported by the response format.
For now we have a logic that adds page image only to first chunk in the LLM prompt and skips it for the other chunks for the same page. But for the retrieval API, I think, this should be configurable by the request parameters.

aidial_rag/retrieval_api.py
Allob: Adobe uses term `page number` here: https://helpx.adobe.com/acrobat/kb/link-html-pdf-page-acrobat.html
and here: https://opensource.adobe.com/pdfembed-sdk-docs/#go-to-a-page
So, `page_number` looks appropriate.

aidial_rag/retrieval_chain.py
Allob: In the current version or RAG, the page image is attached to the most relevant chunk (by image retriever if available, or by the merged results of all the retrievers) when we are building the prompt to the LLM.
But for the retrieval API this step will be configurable. Some clients will want to prepare the prompt the same way the RAG does with minimal changes/code on their side - the same logic of skipping already used images will be added. Some clients will want more control and will get the images for all the chunks. I want the same RetrievalResult to support both options.

aidial_rag/retrieval_api.py
Allob: `url` will not work here, because it is not an url of the chunk, it is an url of the source of this chunk.
`doc_url` is not good also, because the source in not always the whole document, but may be the page of the document. In the future it could some region of the page when more url fragment types will be supported, or it could be the link to the particular paragraph in the html or markdown. I'll fix the description here.

Name `source_url` could work here, but the term `source` is already used not only for the retrieval result, but everywhere in the RAG, including the indexing process and in the prompt we pass to the LLM.
Since you asked for smaller PRs, let's discuss renaming `source` to `source_url` separately and, if agreed, implement in a separate PR?

aidial_rag/retrieval_api.py
Allob: This is the display_name for the source, not for the document.

aidial_rag/qa_chain.py
Allob: Fixed. Now it is passed in chain_input.

aidial_rag/app.py

aidial_rag/app.py
Allob: I did not fully understand what you suggesting here. Hopefully, it is fixed. `chain_input` creation was moved into `_run_rag`.
Now `request_config` parameter is passed to `_run_rag`, and `_run_rag` is responsible to get `chat_chain` and pass it to `generate_answer` as a part of `chain_input`.

aidial_rag/retrieval_chain.py
Allob: Fixed. But I do not like `Tuple[int, int|None]`, so `PageKey` dataclass was added. And `None` is done separately.

aidial_rag/retrieval_chain.py
Allob: Fixed. Now it is `Dict[str, Any]`. However, the `input` dictionary as a a chain parameter is not that typed anyway.

aidial_rag/retrieval_api.py
Allob: Fixed.

aidial_rag/qa_chain.py
```diff
@@ -54,49 +48,43 @@
...
+    retrieval_results: RetrievalResults,
+) -> List[MessageElement]:
+    docs_message: List[MessageElement] = []
     docs_message.append(text_element("<context>"))
-    for i, chunk_metadata in enumerate(chunks_metadatas, start=1):
-        doc_record = doc_records[chunk_metadata["doc_id"]]
-        chunk = doc_record.chunks[chunk_metadata["chunk_id"]]
-
-        attributes = format_attributes(i, chunk.metadata)
+    for i, chunk in enumerate(retrieval_results.chunks, start=1):
```
Allob, line 80: I had to bring this part of the changes into this PR, when I changed the type for the `image_by_page`. And fix the Dict typing for the nearby code.

aidial_rag/qa_chain.py
```diff
@@ -54,49 +48,43 @@
...
 
-INCLUDED_ATTRIBUTES = ["source", "page_number", "title"]
+INCLUDED_ATTRIBUTES = ["page_number", "source", "title"]
 
 
-def format_attributes(i, metadata: dict) -> str:
+def format_attributes(i, metadata: Dict[str, int | str]) -> str:
     attributes = [("id", i)] + [
-        (k, v) for k, v in metadata.items() if k in INCLUDED_ATTRIBUTES
+        (k, metadata[k]) for k in INCLUDED_ATTRIBUTES if k in metadata
```
Allob, line 56: Now the order of the attributes in the LLM prompt is defined by INCLUDED_ATTRIBUTES, not by the order of the fields in the metadata dict.
The order of the fields in the metadata dict was changed after migrating to the RetrievalResults.Chunk, and the cached LLM requests do not match without this change.

aidial_rag/retrieval_api.py
adubovik: > Fox example, for html or markdown we may want to to pass the images embedded in the text.

All the more reason to remove `page` from the field name. 
What is the meaning of a page in HTML/Markdown document?

Page is an instance of a more general concept of **chunk position in a document**.
1. Position in PDF document is defined by page number and optionally bounding box within the page.
2. Position in HTML may be defined by an id of a DOM element.
3. Position in Markdown/Plaintext is defined by (start_row, start_col) + (end_row, end_col).

aidial_rag/qa_chain.py
```diff
@@ -49,49 +48,43 @@
...
-            chunk.metadata.get("page_number"),
-        )
-        if image_key not in attached_images and (
-            image := image_by_page.get(image_key)
-        ):
-            docs_message.append(image_element(image))
-            attached_images.add(image_key)
+        if chunk.page_image_index is not None:
+            image = retrieval_results.images[chunk.page_image_index]
+            if image:
```
adubovik: redundant `if image:`

aidial_rag/retrieval_api.py
```diff
@@ -0,0 +1,66 @@
...
+            description="Human-readable name of the source (the same that Dial RAG displays in the stages).",
+        )
+        text: str | None = Field(
+            default=None,
+            description="Text of the chunk, may be empty, for example, for an image.",
+        )
+        page_number: int | None = Field(
+            default=None,
+            description="The the page number in the document, the chunk belongs to, if applicable. "
+            "Starting from 1, i.e. the same as in the page fragment of the URL.",
```
adubovik: The same style of description for all **index** fields will be great.

adubovik (approved):

Allob:
/deploy-review
>Environment URL: [review-environment](https://chat-ai-dial-rag-pr-27.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/648551)
>Chat E2E tests status: [skipped](https://github.com/epam/ai-dial-ci/actions/runs/16372611156)
>Overlay E2E tests status: [skipped](https://github.com/epam/ai-dial-ci/actions/runs/16372611156)

PR #28, opened 2025-07-16
feat: allow to set index_url outside of index_storage

Allob:
### Description of changes

- index_url is now set outside of the index_storage to allow to set desired index store location in the request parameters in the future
- IndexingTask class was added to specify which index_url correcponds to which attachment_link
- Dial requests were moved from index_storage to DialApiClient to be able to get the bucket_id from Dial outside of the index_storage. Other Dial API requests will be moved to the DialApiClient  in the future

### Checklist

### DIFF
--- a/aidial_rag/app.py
+++ b/aidial_rag/app.py
@@ -23,11 +23,13 @@ from aidial_rag.commands import (
     process_commands,
 )
 from aidial_rag.config_digest import ConfigDigest
+from aidial_rag.dial_api_client import create_dial_api_client
 from aidial_rag.dial_config import DialConfig
 from aidial_rag.document_record import Chunk, DocumentRecord
 from aidial_rag.documents import load_documents
 from aidial_rag.index_record import ChunkMetadata, RetrievalType
-from aidial_rag.index_storage import IndexStorage
+from aidial_rag.index_storage import IndexStorage, link_to_index_url
+from aidial_rag.indexing_task import IndexingTask
 from aidial_rag.qa_chain import generate_answer
 from aidial_rag.repository_digest import (
     RepositoryDigest,
@@ -187,6 +189,19 @@ def create_retriever(
     return retriever
 
 
+def create_indexing_tasks(
+    attachment_links: List[AttachmentLink],
+    dial_api_client,
+) -> List[IndexingTask]:
+    return [
+        IndexingTask(
+            attachment_link=link,
+            index_url=link_to_index_url(link, dial_api_client.bucket_id),
+        )
+        for link in attachment_links
+    ]
+
+
 def get_configuration(request: Request) -> dict:
     if (
         request.custom_fields is None
@@ -291,10 +306,18 @@ class DialRAGApplication(ChatCompletion):
                 get_attachment_links(request_context, messages)
             )
 
+            dial_api_client = await create_dial_api_client(request_context)
+
+            # TODO: Allow to specify desired index URLs in the request
+            indexing_tasks = create_indexing_tasks(
+                attachment_links, dial_api_client
+            )
+
             docs_and_errors = await load_documents(
                 request_context,
-                attachment_links,
+                indexing_tasks,
                 self.index_storage,
+                dial_api_client,
                 config=request_config,
             )
             document_records, loading_errors = process_load_errors(
--- a/aidial_rag/app.py
+++ b/aidial_rag/app.py
@@ -23,7 +23,7 @@ from aidial_rag.commands import (
     process_commands,
 )
 from aidial_rag.config_digest import ConfigDigest
-from aidial_rag.dial_api_client import create_dial_api_client
+from aidial_rag.dial_api_client import DialApiClient, create_dial_api_client
 from aidial_rag.dial_config import DialConfig
 from aidial_rag.document_record import Chunk, DocumentRecord
 from aidial_rag.documents import load_documents
@@ -191,7 +191,7 @@ def create_retriever(
 
 def create_indexing_tasks(
     attachment_links: List[AttachmentLink],
-    dial_api_client,
+    dial_api_client: DialApiClient,
 ) -> List[IndexingTask]:
     return [
         IndexingTask(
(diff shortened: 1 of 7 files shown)


aidial_rag/app.py
```diff
@@ -187,6 +189,19 @@ def stage(retriever, name):
     return retriever
 
 
+def create_indexing_tasks(
+    attachment_links: List[AttachmentLink],
+    dial_api_client,
```
adubovik: `dial_api_client: DialApiClient`

aidial_rag/documents.py
```diff
@@ -226,27 +228,33 @@ def handle_document_processing_error(
 
 async def load_document(
     request_context: RequestContext,
-    attachment_link: AttachmentLink,
+    task: IndexingTask,
     index_storage: IndexStorage,
+    dial_api_client: DialApiClient,
     config: RequestConfig,
 ) -> DocumentRecord:
     with handle_document_processing_error(
```
adubovik, line 235: Nit: `attachment_link = task.attachment_link` to avoid 15 diff lines.

aidial_rag/index_storage.py
```diff
@@ -45,37 +48,43 @@ class IndexStorageConfig(BaseConfig):
...
+# This is treated as a part of an algorithm, not a configuration parameter,
+# because if changed, the old index files will not be found.
+INDEX_PATH_PART_SIZE = 8
+
+
+def link_to_index_url(attachment_link: AttachmentLink, bucket_id: str) -> str:
+    key = hashlib.sha256(attachment_link.dial_link.encode()).hexdigest()
+
+    # split the key into parts to avoid too many files in one directory
+    dir_path = "/".join(
```
adubovik, line 60: `dir_path = "/".join(textwrap.wrap(key, INDEX_PATH_PART_SIZE))`

aidial_rag/index_storage.py
```diff
@@ -45,37 +48,43 @@ class IndexStorageConfig(BaseConfig):
...
+
+    return f"files/{bucket_id}/dial-rag-index/{dir_path}/index.bin"
 
 
 class IndexStorageBackend(ABC):
     @abstractmethod
-    async def load(
-        self, key: str, request_context: RequestContext
-    ) -> bytes | None:
+    async def load(self, url: str, *args, **kwargs) -> bytes | None:
```
adubovik: Var args in the abstract methods is a bizarre choice. 
It **forces** all of the instances of this ABC class to support var args _(somehow)_ too.

Why not move var args to the members of particular class instances?

E.g. 

```diff
class DialFileStorage(IndexStorageBackend):
    def __init__(self, dial_url: str, dial_api_client: DialApiClient):
        self._dial_url = dial_url
        self._dial_base_url = f"{dial_url}/v1/"
+       self._dial_api_client = dial_api_client

    def to_form_data(self, key: str, data: bytes) -> aiohttp.FormData:
        form_data = aiohttp.FormData()
        form_data.add_field(
            "file", io.BytesIO(data), filename=key, content_type=INDEX_MIME_TYPE
        )
        return form_data

    async def load(
-        self, url: str, dial_api_client: DialApiClient
+        self, url: str
    ) -> bytes | None:
        try:
-            return await dial_api_client.get_file(url)
+            return await self._dial_api_client.get_file(url)
        except aiohttp.ClientError as e:
            logger.warning(f"Failed to load index from {url}: {e}")
            return None

    async def store(
-        self, url, data: bytes, dial_api_client: DialApiClient
+        self, url, data: bytes
    ) -> dict:
-        return await dial_api_client.put_file(url, data, INDEX_MIME_TYPE)
+        return await self._dial_api_client.put_file(url, data, INDEX_MIME_TYPE)
```

aidial_rag/index_storage.py
```diff
@@ -86,28 +95,7 @@ class DialFileStorage(IndexStorageBackend):
 
     def __init__(self, dial_url: str):
         self._dial_url = dial_url
```
adubovik: `self._dial_url` and `self._dial_base_url` are unused.

aidial_rag/index_storage.py
```diff
@@ -86,28 +95,7 @@ class DialFileStorage(IndexStorageBackend):
...
-        # split the key into parts to avoid too many files in one directory
-        dir_path = "/".join(
-            key[i : i + dir_part_size]
-            for i in range(0, len(key), dir_part_size)
-        )
-
-        return f"{self._dial_url}/v1/files/{bucket_id}/dial-rag-index/{dir_path}/index.bin"
+        self._dial_base_url = f"{dial_url}/v1/"
 
     def to_form_data(self, key: str, data: bytes) -> aiohttp.FormData:
```
adubovik: `to_form_data` is a dead code.

aidial_rag/index_storage.py
Allob: I would prefer to keep the old logic here. We need just split long string by `INDEX_PATH_PART_SIZE` characters. The code was moved from the `index_url` and we want to keep the behavior the same to work for the already existing indexes.

The `textwrap.wrap` is intended for a different task: "to wrap a paragraph of text". It does not work just on a character level - it takes into account words and whitespaces, has logic to expand tabs and break compound words by hyphens, etc. We do not need all this logic. Also, if any new feature, which may be very useful for paragraph wrapping, will be added to the `textwrap`, I do not want to check if it will change anything for our case - just split the string by N characters.

aidial_rag/app.py

aidial_rag/documents.py
Allob: Fixed.

aidial_rag/index_storage.py
Allob: Fixed.
The issue was that `IndexStorage` was created once at the app level, but the `dial_api_client` is available only during the request, because it encapsulates per-request key and knows about the bucket. Originally, I did not want to touch it in this PR to limit the scope.
Now the `IndexStorage` is created within the request.
In general, the IndexStorageBackend does not work as it was originally intended. I was planning to look at it later when looking at new index format.

aidial_rag/index_storage.py
Allob: Fixed.

aidial_rag/index_storage.py
Allob: Fixed.

aidial_rag/index_storage.py
```diff
@@ -93,30 +92,18 @@ class DialFileStorage(IndexStorageBackend):
...
 
-    async def store(
-        self, url, data: bytes, dial_api_client: DialApiClient
-    ) -> dict:
-        return await dial_api_client.put_file(url, data, INDEX_MIME_TYPE)
+    async def store(self, url, data: bytes) -> dict:
+        return await self._dial_api_client.put_file(url, data, INDEX_MIME_TYPE)
 
 
 class CachedStorage(IndexStorageBackend):
```
adubovik, line 109:
```diff
class CachedStorage(IndexStorageBackend):
-     async def load(self, url: str, *args, **kwargs) -> bytes | None:
+     async def load(self, url: str) -> bytes | None:

-     async def store(self, url, data: bytes, *args, **kwargs) -> dict:
+     async def store(self, url, data: bytes) -> dict:
```

adubovik (approved):

PR #29, opened 2025-07-21
fix: move create_retriever to retrieval_chain

Allob:
### Description of changes

- move create_retriever to retrieval_chain.py
- create_retriever no longer depends on the RequestContext and Choice; it accepts `make_retrieval_stage` as to have a way to report retrieval stage results to the user.
- eval_retriever.py was fixed to use new create_retriever 
- type hints were added for the eval_retriever.py, since I anticipate that it will be requested

### Checklist

### DIFF
--- a/aidial_rag/retrieval_chain.py
+++ b/aidial_rag/retrieval_chain.py
@@ -1,12 +1,16 @@
+import asyncio
 from dataclasses import dataclass
 from itertools import groupby
 from operator import itemgetter
-from typing import Any, Dict, List, Set
+from typing import Any, Callable, Dict, List, Set
 
+from langchain.retrievers import EnsembleRetriever
 from langchain.schema import BaseRetriever, Document
 from langchain.schema.runnable import RunnablePassthrough
+from langchain_core.retrievers import RetrieverLike
 from langchain_core.runnables import Runnable, chain
 
+from aidial_rag.dial_config import DialConfig
 from aidial_rag.document_record import DocumentRecord
 from aidial_rag.image_processor.base64 import pil_image_as_base64
 from aidial_rag.image_processor.extract_pages import (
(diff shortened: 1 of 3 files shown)


aidial_rag/retrieval_chain.py
```diff
@@ -156,10 +168,93 @@ async def create_retrieval_results(
...
 
 
+def _make_retrieval_stage_default(
+    retriever: BaseRetriever, stage_name: str
+) -> BaseRetriever:
+    """Default stage maker that returns the retriever as is."""
+    return retriever
+
+
+def create_retriever(
```
Allob, line 178: This function is mostly moved from app.py. The changes:
- `create_retriever` does not depend on the `Choice` anymore; `make_retrieval_stage` is provided instead.
- `IndexingConfig` is used instead or just `MultimodalIndexConfig`, to hide the retrievers details from the caller.

eval/eval_retriever.py
```diff
@@ -29,42 +31,44 @@
...
         absolute_url=document_link,
         display_name=document_link.split("/")[-1],
     )
 
-    index_config = IndexingConfig(
-        multimodal_index=None,
-        description_index=None,
-    )
-
     doc_record = await load_document_impl(
```
adubovik, line 53: It's disturbing that the source code in the `eval` folder isn't even linted/type-checked.
If the code is neither type-checked, nor integrated into CI, nor regularly executed by the devs themselves in a local setup, it gets quickly outdated and non-functional.

I strongly recommend enabling linting for all source code we maintain in the repo in one of the follow-up PRs.

aidial_rag/retrieval_chain.py
```diff
@@ -156,10 +168,93 @@ async def create_retrieval_results(
...
+    dial_config: DialConfig,
+    document_records: List[DocumentRecord],
+    indexing_config: IndexingConfig,
+    make_retrieval_stage: Callable[
+        [BaseRetriever, str], BaseRetriever
+    ] = _make_retrieval_stage_default,
+) -> BaseRetriever:
+    if not AllDocumentsRetriever.is_within_limit(document_records):
```
adubovik, line 186: Opt: shortcut with `AllDocumentsRetriever` first, to reduce the nesting and streamline the code.

```python
if AllDocumentsRetriever.is_within_limit(document_records):
    return make_retrieval_stage(
            AllDocumentsRetriever.from_doc_records(document_records),
            "All documents",
        )

...
return make_retrieval_stage(
    EnsembleRetriever(
        retrievers=retrievers,
        weights=weights,
    ),
    "Combined search",
)
```

adubovik (approved):

aidial_rag/retrieval_chain.py
Allob: I agree that suggested change will improve the code. 
But in this PR the code of the function moved here mostly unchanged. I would prefer to implement the suggested change separately to keep the code easier to merge with upcoming PRs.

eval/eval_retriever.py
Allob: Mark this as resolved since you suggest to address it in the follow-up PRs.

PR #30, opened 2025-07-21
chore: use aidial-rag-eval from public repository

Allob:
### Description of changes

- update dependency to use aidial-rag-eval from public repository

### Checklist


pyproject.toml
```diff
@@ -35,7 +35,7 @@ python-magic-bin = {version = "^0.4.14", platform = "win32"}
 nltk = "^3.9.1"
 humanfriendly = "10.0"
 pyinstrument = "4.6.1"
-more-itertools = "10.1.0"
+more-itertools = "^10.1.0"
```
Allob, line 37: aidial-rag-eval requires more-itertools>=10.5.0

adubovik: Why not `more-itertools = "^10.5.0"` then?

`more-itertools = "^10.1.0"` implies among other things that the project is compatible with `more-itertools==10.1.0`, but in fact **it isn't**. It's confusing.

adubovik: Moreover, declaring the version ranges for the dependencies makes little sense for an **application**, since ultimately the application is tested against and shipped with dependency versions pinned in the **poetry.lock** file.

Compare it with the dial-sdk **library**, which we actually test against different sets of dependency versions.

Allob: The code in this project has not changed. It is still compatible with `more-itertools==10.1.0`. One of the dependencies asks for `more-itertools>=10.5.0`.

If one of the dependencies will require `10.6.0` or newer, I do not want to change this line for every such change. `poetry.lock` should do it for me.
Btw, the actual version in `poetry.lock` is `10.7.0`.

> Moreover, declaring the version ranges for the dependencies makes little sense for an application, since ultimately the application is tested against and shipped with dependency versions pinned in the poetry.lock file.

Yes, the application is tested against and shipped (for the docker image, not wheel) with dependency versions pinned in the `poetry.lock` file.

But the version range still make sense here. I'm not sure if you suggesting the exact version instead of range, or no version at all, but options are worse than this range.
- range vs no version - the code in this project most likely will not be compatible with the version `9.x.x`. And new `11.0.0` may introduce the incompatibility which would require the changes in the code. The range allows the poetry to take it into account when resolving dependencies and fixing the version in the `poetry.lock` file.
- range vs fixed version - having the range will allow tools to know if it is possible to update the version in the `poetry.lock` file if needed. For example, the dependabot may bring some security fix with `10.x.x` version, but not `11.x.x`, since it will not be compatible. Or some library update may bring `10.8.0` or higher, which still should be compatible with the project code.

> Compare it with the dial-sdk library, which we actually test against different sets of dependency versions.

I do not understand what you wanted to show me with `dial-sdk` library. I can see that, for example, for pydantic:
- in pyproject.toml there is a range - `pydantic (>=1.10.17,<3)`
- in `poetry.lock` the version is `2.8.0`
- but in the noxfile versions are `["1.10.17", "2.8.2"]`
I.e. version `2.8.0`, which is specified in `poetry.lock`, is not used in the tests. Versions from `2.0.x` to `2.7.x` also are not used in the tests. So, I do not understand what point you wanted to explain me with the dial-sdk library example.

Allob:
/deploy-review
>Environment URL: [review-environment](https://chat-ai-dial-rag-pr-30.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/669905)

pyproject.toml
```diff
@@ -29,13 +29,12 @@ rank-bm25 = "0.2.2"
 sentence-transformers = {version = "3.3.1", extras = ["openvino"]}
 tiktoken = "^0.7.0"
 cachetools = "5.3.1"
-unstructured = {version = "0.16.14", extras = ["doc", "docx", "pdf", "ppt", "pptx", "xlsx"]}
-libmagic = "1.0"
+unstructured = {version = "0.16.16", extras = ["doc", "docx", "pdf", "ppt", "pptx", "xlsx"]}
```
Allob, line 32: unstructured was updated to 0.16.16 to remove dependency on ndjson.

adubovik (approved):

Allob:
/deploy-review
>GitHub actions run: [17551131160](https://github.com/epam/ai-dial-ci/actions/runs/17551131160)
>Environment URL: [review-environment](https://chat-ai-dial-rag-pr-30.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/680800)

PR #36, opened 2025-07-28
fix: fix hanging tests

Allob:
### Description of changes

Fixes hanging tests for the application.
There is an issue that the tests for the application (with fastapi.testclient) may hang. For example, this run: https://github.com/epam/ai-dial-rag/actions/runs/16473749024/job/46569649203?pr=34

The reason is following:
- We use `langchain_unstructured.UnstructuredLoader` to parse the document using the open source unstructured library.
- `partition_via_api` is `False` by default - we do not use unstructured API, but `langchain_unstructured.UnstructuredLoader` creates  `UnstructuredClient` anyway: https://github.com/langchain-ai/langchain-unstructured/blob/main/libs/unstructured/langchain_unstructured/document_loaders.py#L138
- The `UnstructuredClient` registers `close_clients` as `weakref.finalize` which could be called by garbage collector at any time: https://github.com/Unstructured-IO/unstructured-python-client/blob/9576035a9c9f7eef97da90db177bf057e5f62471/src/unstructured_client/sdk.py#L107
- The `close_clients` uses `ThreadPoolExecutor.submit` to schedule the `aclose` for the client  https://github.com/Unstructured-IO/unstructured-python-client/blob/9576035a9c9f7eef97da90db177bf057e5f62471/src/unstructured_client/httpclient.py#L129
- If `close_clients` happen to be called during `ThreadPoolExecutor.submit` itself, it could block the thread of the lock in ThreadPoolExecutor and cause a deadlock. See screenshot:
Screenshot 2025-07-28 212315

This PR changes `document_loaders.py` to avoid creating/destroying `UnstructuredClient` to avoid this issue.

Note: the `close_clients` code has been changed several times since the version we use and is different now: https://github.com/Unstructured-IO/unstructured-python-client/blob/main/src/unstructured_client/httpclient.py#L120 . But, I think, it would be better to avoid the issue at all, since we do not use the `UnstructuredClient` anyway.

### Checklist

### DIFF
--- a/aidial_rag/document_loaders.py
+++ b/aidial_rag/document_loaders.py
@@ -14,6 +14,7 @@ from langchain_unstructured import UnstructuredLoader
 from pdf2image.exceptions import PDFInfoNotInstalledError
 from pydantic import ByteSize, Field
 from unstructured.file_utils.model import FileType
+from unstructured_client import UnstructuredClient
 from unstructured_pytesseract.pytesseract import TesseractNotFoundError
 
 from aidial_rag.attachment_link import AttachmentLink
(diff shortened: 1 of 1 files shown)


adubovik (dismissed):

aidial_rag/document_loaders.py
```diff
@@ -29,6 +30,17 @@
 from aidial_rag.resources.cpu_pools import run_in_indexing_cpu_pool
 from aidial_rag.utils import format_size, get_bytes_length, timed_block
 
+# Create a stub for UnstructuredClient to avoid creating/destroying it in every request.
```
adubovik: I think it's important to emphasise in the comment that the stub must be a **global object** so that GC doesn't collect it at some unfortunate moment, leading to a deadlock.

Otherwise, someone reading this comment two years from now will be tempted to inline the object in the `get_document_chunks` function.

Allob: Fixed. Added "as a global object" to the comment.
However, if someone in two years will put in into the App object, or in a class-level field - it would also be acceptable. The key point is an `UnstructuredClient` lifetime, which is mentioned in the comment.

adubovik (approved):

PR #40, opened 2025-08-01
feat: add indexing request

Allob:
### Description of changes

Introduce indexing requests:
- `request.type == "indexing"` was added, which makes RAG to perform indexing only and stop before retrieval step.
- indexing requests return the index files as attachments with `reference_url` pointing to the original file
- indexing errors are reported in attachment with type `application/x.aidial-rag.indexing-response+json`
- index files may be passed as an attachments (with `reference_url` pointing to the original file) to a retrieval or rag request to use pre-calculated indexes

### Checklist


Allob:
/deploy-review
>Environment URL: [review-environment](https://chat-ai-dial-rag-pr-40.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/658705)

aidial_rag/index_storage.py
```diff
@@ -46,6 +47,13 @@ class IndexStorageConfig(BaseConfig):
 
 INDEX_MIME_TYPE = "application/x.aidial-rag-index.v0"
 
+INDEX_MIME_TYPES_REGEX = re.compile(r"^application/x\.aidial-rag-index\..*$")
```
adubovik: Let's put ourselves in the place of a client of the Retrieval/Indexing RAG API.
Which modules should I read to learn about all the data types relevant to the API? 
My understanding is that it's `*_api` modules, isn't?

So maybe move `INDEX_MIME_TYPE` to `indexing_api` module, since it's a part of the client-facing indexing API.

aidial_rag/app.py
```diff
@@ -98,51 +106,71 @@ def doc_to_attach(
...
+                index_attachments.get(link.dial_link)
+                or link_to_index_url(link, dial_api_client.bucket_id)
+            ),
         )
         for link in attachment_links
+        if not _is_rag_index(link)
     ]
 
 
+def _return_indexing_results(
```
adubovik: Calling a function `_return_something` is strange, because it's ambiguous, since a function always returns something (even if None). 

Maybe `_add/dump/save/display/report_indexing_results`?

aidial_rag/app.py
```diff
@@ -98,51 +106,71 @@ def doc_to_attach(
...
+    choice: Choice,
+    indexing_results: List[DocumentIndexingResult],
+) -> None:
+    attachments = create_indexing_results_attachments(indexing_results)
+    for attachment in attachments:
+        choice.add_attachment(attachment)
+    return
```
adubovik: Redundant `return`

aidial_rag/indexing_results.py
```diff
@@ -0,0 +1,111 @@
...
+from aidial_sdk import HTTPException
+from pydantic import BaseModel, ConfigDict, model_validator
+from requests.exceptions import Timeout
+
+from aidial_rag.document_record import DocumentRecord
+from aidial_rag.errors import DocumentProcessingError
+from aidial_rag.indexing_task import IndexingTask
+
+
+class DocumentIndexingResult(BaseModel):
```
adubovik, line 12: There are **a lot** of places where we check that the indexing task was either successful or failed with an exception.

I would recommend splitting the class into two for these two cases to make the disjoint nature of the outcome explicit.

```python
class DocumentIndexingResult(BaseModel):
    task: IndexingTask

class DocumentIndexingSuccess(DocumentIndexingResult):
    doc_record: DocumentRecord

class DocumentIndexingFailure(DocumentIndexingResult):
    exception: Exception

    # Maybe cached property `leaf_exceptions`?
    def iter_leaf_exceptions(self) -> Generator[BaseException, None, None]:
        ...

class DocumentIndexingResults(BaseModel):
    successes: List[DocumentIndexingSuccess]
    failures: List[DocumentIndexingFailure]

def has_document_loading_errors(results: DocumentIndexingResults) -> bool:
    ...

def create_index_attachment(
    indexing_result: DocumentIndexingSuccess,
) -> Attachment:
    ...

def create_indexing_response(
    indexing_results: List[DocumentIndexingFailure],
) -> IndexingResponse:
    ...

def format_document_loading_errors(
    indexing_results: List[DocumentIndexingFailure],
) -> str:
    ...

def create_document_loading_exception(
    indexing_results: List[DocumentIndexingFailure],
) -> HTTPException:
    ...
```

tests/test_app_indexing.py
```diff
@@ -0,0 +1,295 @@
...
+    assert indexing_response.status_code == 200
+    indexing_response_attachments = indexing_response.json()["choices"][0][
+        "message"
+    ]["custom_content"]["attachments"]
+
+    # alps_wiki.html should be indexed successfully
+    index_attachments = [
+        attachment
+        for attachment in indexing_response_attachments
+        if attachment["type"] == "application/x.aidial-rag-index.v0"
```
adubovik: Maybe replace `application/x.aidial-rag-index.v0` with `INDEX_MIME_TYPE` here and in other places throughout the tests.

aidial_rag/configuration_endpoint.py
```diff
@@ -20,7 +20,7 @@ class RequestType(StrEnum):
 
     RAG = "rag"
     RETRIEVAL = "retrieval"
-    # TODO: Add INDEXING request type
+    INDEXING = "indexing"
```
adubovik, line 23: It would be great to document this new configuration option in README, since it is part of the public DIAL RAG API.

aidial_rag/documents.py
```diff
@@ -282,16 +283,38 @@ async def load_document(
...
-) -> List[DocumentRecord | BaseException]:
+) -> List[DocumentIndexingResult]:
+    # TODO: Rewrite this function using TaskGroup to cancel all tasks if one of them fails
+    # if ignore_document_loading_errors is not set in the config
     return await asyncio.gather(
         *[
-            load_document(request_context, task, index_storage, config)
+            load_document_task(request_context, task, index_storage, config)
             for task in tasks
         ],
```
adubovik, line 324: 1. `return_exceptions=True` is used to catch a very broad range of exceptions - `BaseException`s. Now only the `DocumentProcessingError` errors are being caught. Any other exception will blow the whole batch of document-processing tasks. Is it intended?

2. If yes, then (a) `DocumentIndexingResult.exception` has a broader exception type than it actually takes:  `Exception | None` instead of `DocumentProcessingError | None`; (b) `_iter_leaf_exceptions` method is too generic too, since it traverses `BaseException`.

aidial_rag/indexing_api.py
```diff
@@ -0,0 +1,87 @@
...
+    for result in indexing_results:
+        if result.exception is None:
+            index_attachments.append(create_index_attachment(result))
+
+    indexing_response = create_indexing_response(indexing_results)
+    index_attachments.append(
+        Attachment(
+            title="Indexing results",
+            type=indexing_response.CONTENT_TYPE,
+            data=indexing_response.model_dump_json(indent=2),
```
adubovik: `indent=2` - should we add redundant spaces to the HTTP JSON response body?

aidial_rag/index_storage.py
Allob: Fixed.

aidial_rag/app.py
Allob: Changed to `_add_indexing_results(choice, indexing_results)`

aidial_rag/app.py

tests/test_app_indexing.py
Allob: Fixed.

aidial_rag/configuration_endpoint.py
Allob: I agree. I was planning to write an md file and add a py-code with a usage example in a separate PR.

aidial_rag/documents.py
Allob: 1. The plan was that normal document processing exceptions should not fly here. They should be caught in the `load_document_task` and returned in the `DocumentIndexingResult`. So, if any exception flies out of `load_document_task` (like `KeyboardInterrupt` or `SystemExit`) - it means that it is something abnormal and it should fail the whole process.
2. Actually, `DocumentProcessingError` is somewhat redundant and could be removed later (with some changes in exception logging). Partially fixed: I removed `DocumentProcessingError` from the `DocumentIndexingResult`. It is limited by `documents.py` only and may be removed later.

aidial_rag/indexing_api.py
Allob: Fixed.

aidial_rag/indexing_results.py
Allob: I agree with DocumentIndexingSuccess and DocumentIndexingFailure, because it allows to remove custom validator.
And I agree that `format_document_loading_errors` and `create_document_loading_exception` could just get a list of failures.
But I do not like `DocumentIndexingResults` with separate `successes` and `failures` lists, because it makes harder to check the result for the particular document. And if `asyncio.gather` will be rewritted to `TaskGroup` and we will get the third state - `cancelled` - we will need to look into a 3 separate lists to know the result for a particular document.
Also, I do not agree with the suggestions `create_indexing_response`. The fact that `create_indexing_response` needs just write errors to the response json - is an implementation detail of the `create_indexing_response`. The format could be changed to also write `ok` or some statistics for successful documents.
Partially fixed.

aidial_rag/indexing_results.py
```diff
@@ -0,0 +1,114 @@
...
+    """Result of a failed indexing operation."""
+
+    exception: Exception
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    def iter_leaf_exceptions(self) -> Generator[BaseException, None, None]:
+        """Iterate over leaf exceptions in the result."""
+        if self.exception is None:
+            return
```
adubovik: Redundant check

Allob: Fixed. Thanks.

adubovik (approved):

PR #41, opened 2025-08-05
fix: check index path for dial-rag bucket

Allob:
### Description of changes

If the index path provided by the used is pointing to the dial-rag bucket, check that the path matched with the expected path for the index to avoid overwriting the indexes of another documents.

### Checklist

### DIFF
--- a/aidial_rag/index_mime_type.py
+++ b/aidial_rag/index_mime_type.py
@@ -1,0 +1,10 @@
+import re
+
+INDEX_MIME_TYPE = "application/x.aidial-rag-index.v0"
+INDEX_MIME_TYPES_REGEX = re.compile(r"^application/x\.aidial-rag-index\..*$")
+
+
+assert INDEX_MIME_TYPES_REGEX.match(INDEX_MIME_TYPE), (
+    f"Invalid INDEX_MIME_TYPE: {INDEX_MIME_TYPE}. "
+    f"It should match the regex {INDEX_MIME_TYPES_REGEX.pattern}."
+)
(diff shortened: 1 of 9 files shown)


aidial_rag/index_mime_type.py
```diff
@@ -0,0 +1,10 @@
+import re
+
+INDEX_MIME_TYPE = "application/x.aidial-rag-index.v0"
```
Allob, line 3: Moved to separate file to avoid circular dependencies.

aidial_rag/indexing_task.py
```diff
@@ -10,3 +16,86 @@ class IndexingTask(BaseModel):
...
+    expected_index_url = link_to_index_url(
+        task.attachment_link, dial_api_client.bucket_id
+    )
+    if index_url != expected_index_url:
+        raise InvalidAttachmentError(
+            f"Index URL {index_url} does not match the expected index path {expected_index_url}."
+        )
+
+
+def create_indexing_tasks(
```
Allob, line 79: Moved from `app.py`.

Originally, I was planning to add `index_url` validation to this function. But we could have invalid index url for multiple documents, and we need some error reporting mechanism for this. So, the validation is done in load_document to se already existing error handling for multiple documents.

So, this move from `app.py` is not needed for the final version of the changes. But I kept it here, because having less stuff in `app.py` is better in general.

adubovik (approved):

Allob:
/deploy-review
>Environment URL: [review-environment](https://chat-ai-dial-rag-pr-41.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/660485)

PR #42, opened 2025-08-06
feat: add documentation and examples for indexing and retrieval api

Allob:
### Description of changes

Add notebooks with usage examples for indexing and retrieval api.

### Checklist


docs/api.md
```diff
@@ -0,0 +1,250 @@
+# Dial RAG API Documentation
+
+The Dial RAG API allows you to customize what parts of the RAG pipeline you want to use. In general, the Dial RAG pipeline consists of the following steps:
+- **Indexing** - the process of parsing the documents and creating an index of the document chunks. By default, Dial RAG handles the storage of the document indexes.
+- **Retrieval** - the process of retrieving relevant chunks of documents based on the user query. By default, Dial RAG automatically finds the index for the documents in the request and runs the indexing if the documents were not indexed before.
+- **Response Generation** - the process of generating a response based on the retrieved document chunks and the user query.
```
adubovik: `Response` in `Response Generation` is ambiguous, since each of these 3 steps results in some sort of a response.
So how is the response generation different from the other two steps?

Maybe rename to simply `Generation` - after all RAG is Retrieval-augmented Generation, not Retrieval-augmented Response Generation.

docs/api.md
```diff
@@ -0,0 +1,250 @@
+# Dial RAG API Documentation
```
adubovik: Find-and-replace `Dial` => `DIAL`.

Likewise in Python notebooks.

noxfile.py
```diff
@@ -69,6 +75,24 @@ def update_docs(session):
...
+        "python",
+        "./generate_json_schema.py",
+        "aidial_rag.indexing_api.IndexingResponse",
+        "docs/indexing_response.generated.schema.json",
+    )
+    session.run(
+        "python",
+        "./generate_json_schema.py",
+        "aidial_rag.configuration_endpoint.Configuration",
+        "docs/configuration.generated.schema.json",
```
adubovik, line 94: `generated` in `configuration.generated.schema.json` tells how the file came to be.
I'm not sure it's something a DIAL RAG client would care about. 

People mostly care about **what the thing is**, not how it was obtained.
The config might have been manually written, but I don't really care, as soon as it is what it claims to be - the schema of application configuration.

I suggest renaming to `configuration.schema.json`, likewise other files.

docs/api.md
```diff
@@ -0,0 +1,250 @@
...
+
+Additionally, the Dial RAG uses the [`custom_fields.configuration`](https://dialx.ai/dial_api#operation/configurationDeployment) field of the Unified API to configure the RAG pipeline.
+
+You can look at the autogenerated [Dial RAG Configuration schema](./configuration.generated.schema.json) to see all other the available configuration options for the Dial RAG pipeline.
+
+## Indexing and Retrieval API
+
+Indexing and Retrieval API allows you to use the parts of the Dial RAG pipeline. It is based on the Dial Unified API and works via the same `/chat/completions` endpoint.
+To run the Indexing or Retrieval part of the RAG pipeline, you need to set the `custom_fields.configuration.request.type` field in the request body. The Dial RAG supports the following request types:
+- **`rag`** - used by default if no type is specified. Runs the full RAG pipeline, including indexing (if documents were not indexed before), retrieval, and response generation.
```
adubovik: 1/ Usually it helps to sort the list items in the order of execution _(even though some steps could be skipped due to caching)_: indexing, retrieval, rag
2/ Maybe rename rag to `generation`, since in the overview section it's already referred to as `Response Generation` ?

docs/api.md
```diff
@@ -0,0 +1,250 @@
...
+}
+```
+The index attachment should use `url` field to point to the index file in the Dial storage. The `data` field for the embedded file content is not supported.
+The `reference_url` field is used to point to the document the index file is based on, which is useful for the retrieval requests to find the original document. The `reference_url` is required for the index attachments.
+
+
+### Indexing Request
+
+The indexing request is a `/chat/completion` request with `custom_fields.configuration.request.type` field to `indexing` in the request body.
+You should attach the documents to be indexed as a list of index attachments in the `custom_content.attachments` field.
```
adubovik: A hyperlink to the section about index attachments will help: `[index attachments](#index-attachments)`

docs/api.md
```diff
@@ -0,0 +1,250 @@
...
+        ]
+    }
+}
+```
+
+Additionally, the response will contain the [Stages](https://docs.dialx.ai/tutorials/developers/chat/chat-objects#stages) with the progress of the indexing process in the same way as the regular Dial RAG response. The stages will be dynamically updated if the request is run with the `stream` parameter set to `true`.
+
+Since the desired index file location was not specified in the request, the Dial RAG will store the index files in the Dial RAG bucket with a path based on the document URL. You can use the `reference_url` field of the index attachment to find the original document. You can use the `url` field to access the index file in the Dial storage to copy it to your own storage if needed.
+
+**Note**: the internal structure of the index files is not documented and is not intended to be used directly. The index files are meant to be read by the Dial RAG API. You should use the Dial RAG API to retrieve the indexed documents or to run the retrieval requests.
```
adubovik: You can use the special Github syntax for notes to make it stand out:

> [!NOTE]  
> The internal structure of the index files is not documented and is not intended to be used directly. The index files are meant to be read by the Dial RAG API. You should use the Dial RAG API to retrieve the indexed documents or to run the retrieval requests.

...

> [!NOTE]
> The Dial RAG will not allow you to specify the custom index location within the Dial RAG bucket, because the paths within the Dial RAG bucket are managed by the Dial RAG itself, and we want to avoid conflicts with the existing index files.

docs/retrieval_api_example.ipynb
```diff
@@ -0,0 +1,490 @@
...
+   "id": "3cbd7ad8",
+   "metadata": {},
+   "source": [
+    "# Retrieval API Example\n",
+    "\n",
+    "\n",
+    "This notebook demonstrates how to use the Retrieval API of the Dial RAG application.\n",
+    "\n",
+    "\n",
+    "The retrieval API allows you to retrieve relevant information based on user queries without running the full RAG pipeline. It is useful if the customization of the RAG pipeline by the [configuration](https://dialx.ai/dial_api#operation/configurationDeployment) is not enough for your use case. For example, you may want to use LLM with structured output, or [tool calling](https://platform.openai.com/docs/guides/tool-calling), or get the [Log probabilities](https://platform.openai.com/docs/api-reference/chat/create#chat_create-logprobs) of the tokens in the response. In this case, you can use the retrieval API to get the relevant documents and then run your custom LLM pipeline on top of them.\n",
```
adubovik: https://platform.openai.com/docs/guides/tool-calling is 404, use
https://platform.openai.com/docs/guides/function-calling

And the feature is generally known as `function calling`.

docs/indexing_api_example.ipynb
```diff
@@ -0,0 +1,502 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "90b4e061",
+   "metadata": {},
+   "source": [
+    "# Indexing API Example\n",
+    "\n",
+    "This notebook demonstrates how to use the Indexing API of the Dial RAG application.\n",
+    "\n",
+    "Dial RAG can automatically run indexing for a new documents it sees in RAG or retrieval requests. However, you can also manually index documents using the Indexing API. It may be useful if you know the document in advance and do not want tp wait for the automatic indexing during the first request, or if you want to attribute the costs of the indexing to a separate Dial API-Key. Also, the indexing API allows you to manually manage the path to the index storage, which can be useful if you have some custom pipeline which involves copying/moving the documents after indexing.\n",
```
adubovik: typo `do not want tp wait`

docs/indexing_api_example.ipynb
```diff
@@ -0,0 +1,502 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "0c14f6ef",
+   "metadata": {},
+   "source": [
+    "## Indexing\n",
+    "\n",
+    "For this example we will use two files from the test data:\n",
+    "- alps_wiki.pdf - pdf file with wikipedia article about the Alps. This file will be successfully indexed by the Dial RAG and will be used to demonstrate how to use it in Dial RAG requests.\n",
+    "- test_file.csv - the unsupported CSV file with some test data. This CSV file has different number of columns in each row and cannot be parsed as a single table, which is currently not supported by the Dial RAG. This file will help to demonstrate error handling.\n",
```
adubovik: Maybe worth renaming the file to `invalid_csv_file.csv`

generate_json_schema.py
```diff
@@ -0,0 +1,36 @@
+import importlib
```
adubovik, line 1: Maybe worth moving `generate_json_schema.py` and `download_model.py` to a dedicated top-level folder `scripts`.

noxfile.py
Allob: I would like to keep `.generated.` as a marker for someone who would want to edit the schema manually in the future.

docs/api.md
Allob: `rag` here means the full pipeline, not just the generation stage.

The `generation` stage is also theoretically possible here. It would take retrieval results as input and return LLM response. But we are not planning to implement it for now.

generate_json_schema.py
Allob: I agree with this. But I think it should be done in a separate PR, because it will affect the docker image building. I.e. it is not just documentation.

docs/indexing_api_example.ipynb
Allob: This csv file is currently unsupported by DIAL RAG, but it is not invalid CSV itself.

There are two different flavors of CSV files:
- table-like CSV - the file where each column corresponds to some value of the row. This file should have the same number of columns in each row. Such file could be read by, for example, pandas. This type of CSV is supported by the unstructured library and could be read by DIAL RAG.
- Excel sheet-like CSV - the file in this case is just a canvas to put some strings or numbers on it. The data is this file could form some tables, but only visually, because some numbers or strings are places near one another. This type of CSV currently is not supported by the unstructured library and could not be read by DIAL RAG.

docs/api.md
Allob: Fixed. Renamed to `Generation`.

docs/api.md
Allob: Fixed. This change is in a separate commit to make other fixes easier to see.

docs/api.md
Allob: Fixed. Thanks.
It should be just `attachments`, not `index attachments` here. The link was added to `index attachments`link in the next sentence.

docs/api.md
Allob: Fixed. Thanks.

docs/retrieval_api_example.ipynb
Allob: Fixed. Thanks.

docs/indexing_api_example.ipynb
Allob: Fixed.

adubovik (approved):

PR #46, opened 2025-08-22
fix: update rag-eval dependency and license issues in ort file

paknikolai:
### Description of changes

- updated version of aidial-rag-eval

- created .ort.yml file to resolve issues with ort scanner:

  - problematic licenses found in ort were checked and added to resolutions with comment with proper license
 - unstructured version was downgraded because newer version had dependency on ndjson which had GPL license
 - removed libmagic dependency from python since unstructured uses python-magic inside, but installing lib inside the docker image is needed so python-magic can call system lib
 - updated tests caches
- updated openvino to next version which has bug fix for mac, without it importing openvino caused error
### Checklist

- [ ] Title of the pull request follows [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/)


Allob:
This PR is outdated. These changes implemented in https://github.com/epam/ai-dial-rag/pull/30

PR #48, opened 2025-08-29
feat: add temperature parameter to support gpt-5 models

Allob:
### Description of changes

Gpt-5 models require temperature to be set to 1 to work. Adding the `temperature` parameter to support gpt-5 models.

### Checklist


aidial_rag/llm.py
```diff
@@ -27,6 +27,14 @@ class LlmConfig(BaseConfig):
         default=2,
         description="Sets the number of retries to send the request to the LLM.",
     )
+    temperature: float = Field(
+        default=0.0,
```
adubovik, line 31: Is it intended that the default is 0.0? If so, it's quite a major change, since LLM models usually don't have zero temperature by default. Certainly not GPT, whose temperature is 1.0 by default and ranges from 0.0 to 2.0 _(not 1.0, mind you)_.

https://platform.openai.com/docs/api-reference/chat/create#chat_create-temperature
https://platform.openai.com/docs/api-reference/responses/create#responses_create-temperature

Allob: You can see below, that the value 0 was used in the RAG before:
```diff
-        temperature=0,
+        temperature=llm_config.temperature,
```
So, default value is set 0 to preserve the old behavior in case if the parameter is not set (i.e. in the most cases, except the experiments with gpt-5).

adubovik (approved):

PR #58, opened 2025-09-16
feat: support migration for index formats 11 and 12

Allob:
### Description of changes

- add support of index migration from formats 11 and 12

### Checklist


tests/test_doc_record.py
```diff
@@ -32,64 +24,33 @@ def local_server():
...
-
+# To get an index file, make sure that the DIAL RAG is started with
+# DIAL_RAG__INDEX_STORAGE__USE_DIAL_FILE_STORAGE=True
+# and download the index file from the file storage.
 @pytest.mark.parametrize(
     "index_file",
     [
+        "doc_record_0.22.0.bin",  # format 11
+        "doc_record_0.33.0.bin",  # format 12
         "doc_record_0.34.0rc0.bin",  # before modification_metadata
```
Allob: doc_record_0.34.0rc0.bin file was updated to include fully-featured index with image pipeline as well. The index from the `test_prepare_index` has limited features. It is better to run the full rag to update index, if needed in the future.

aidial_rag/index_storage.py
```diff
@@ -126,33 +120,17 @@ def __init__(
...
-    ) -> DocumentRecord:
+    ) -> bytes:
         doc_record_bytes = await self._storage.load(task.index_url)
         if doc_record_bytes is None:
             raise IndexMissingError()
-        try:
-            doc_record = DocumentRecord.from_bytes(
-                doc_record_bytes, **SERIALIZATION_CONFIG
-            )
-            if doc_record.format_version != FORMAT_VERSION:
```
Allob, line 138: deserialization was moved from the index_storage.py to the document.py and IndexStorage now returns `bytes`, because now we need the logic to migrate the old formats and store the new format back to the storage.

aidial_rag/documents.py
```diff
@@ -274,29 +354,12 @@ async def load_document(
                     logger.warning(f"Failed to load index: {e}")
 
         if doc_record is None:
-            with timed_stage(
```
Allob, line 277: `_load_index_stage`, `_process_document_stage` and `_store_index_stage` were extracted from here, since we are getting more and more if-s here to decide when we need to reindex or migrate and index. And more if-s (like etags check) will be added here in the future.

aidial_rag/document_record_migration.py
```diff
@@ -0,0 +1,204 @@
...
+
+    @staticmethod
+    def _migrate_index_settings(
+        old_settings: Format11._IndexSettings,
+    ) -> IndexSettings:
+        # Get default value for all the fields that not supported in old format
+        index_settings = IndexingConfig().collect_fields_that_rebuild_index()
+
+        def _delete_key(d: dict, key: str):
+            if key in d:
```
adubovik: nit: `d.pop(key, None)`

aidial_rag/documents.py
```diff
@@ -248,20 +318,30 @@ async def load_document(
...
-                        load_stage.content_stream, doc_record.chunks
+                doc_record = await _load_index_stage(
+                    request_context, index_storage, task, index_settings
+                )
+                if (
+                    doc_record.format_version is None
+                    or doc_record.format_version < MIN_FORMAT_VERSION
+                ):
+                    raise IndexIncompatibleError(
+                        f"Index format version mismatch: {doc_record.format_version}"
```
adubovik: A mismatch usually involves two things. What is mismatching with what?

tests/test_doc_record.py
adubovik: Maybe not only test the **old** versions of the doc_record, but **all** versions that are currently supported?

And also make explicit the version number in the file names?

```python
[
   "doc_record_v11_0.22.0.bin", 
   "doc_record_v12_0.33.0.bin", 
   "doc_record_v13_without_metadata_0.34.0rc0.bin",
   "doc_record_v13_with_metadata_0.34.0rc0.bin", 
]
```

Also, why do we need to include the package versions in the filenames?

aidial_rag/configuration_endpoint.py
```diff
@@ -46,6 +46,11 @@ class ApiRequest(BaseModel):
         description="Enable indexing of the documents if the index is missing or is not compatible with retrieval.",
     )
 
+    save_index_on_migration: bool = Field(
```
adubovik, line 49: The idea of index migration without any visible side effects _(such as saving the index)_ is kind of pointless from the user's perspective. So we could as well not introduce this idea in the API in the first place.

So I believe `migrate_index` is a more user-friendly name for this option.

aidial_rag/configuration_endpoint.py
```diff
@@ -46,6 +46,11 @@ class ApiRequest(BaseModel):
...
+        default=True,
+        description="Allows to save the migrated index back to the storage if the index was migrated to the current format on load.",
```
adubovik: Maybe worth clarifying what the index is being migrated **from**:

`Allows to save the migrated index back to the storage if the index was migrated from an older index format to the latest one on load.`

tests/test_doc_record.py
Allob: > Maybe not only test the old versions of the doc_record, but all versions that are currently supported?

I can add this, but I'm not worried about the latest version, because it is used in the other tests.

> Also, why do we need to include the package versions in the filenames?

I think that the package is better indicator for the binary index version than the format version:
1. We may have backwards-compatible changes which do not break the format version (like adding `modification_metadata`). I'll hope to see more backwards compatible changes in the future instead of breaking the format compatibility.
2. There could be some other changes like library updates or changes in the configuration or models. These changes should not break the format and should be compatible, but we still may want to test them to make sure. I.e. it looks reasonable to have here several indexes with the same format version, but produced by different versions of the ai-dial-rag. So, ai-dial-rag version looks like a reasonable way to indicate it.

> And also make explicit the version number in the file names?

Ok, I could add it. But I would prefer to but the rag version first, because it is more important as a differentiator:
```
[
   "doc_record_0.22.0_format11.bin", 
   "doc_record_0.33.0_format12.bin", 
   "doc_record_0.34.0rc_format13_without_modification_metadata.bin",
   "doc_record_0.34.0rc_format13_with_modification_metadata.bin",
]
```

aidial_rag/configuration_endpoint.py
Allob: I think I do not fully understand what do you suggest for the `migrate_index`.
If the `migrate_index` is True, I assume, we have to migrate an old index to the new format and save it to the storage.
But if the `migrate_index` is False, should we try to migrate the format on load, but without saving? Or should we do not try to migrate it and throw an error?

Saving index back to the storage may not be possible if an index is managed externally. If some external application, which uses indexing/retrieving api, has copied/moved an index to a new location, the DIAL RAG may not have a write permissions to store the migrated index. An attempt to save an index without write permissions will fail.
I.e. there could be cases when the direct effect of an index migration _(such as being able to work with an old index)_ may still be useful, even if a side-effect _(such as saving the index)_ is disabled.

Also, I have at least one use case for the tests in the `tests/test_app_old_indexes.py`. I want to migrate old indexes on load, but I do not want to write the new versions back to the storage and overwrite my test data.

aidial_rag/configuration_endpoint.py
Allob: Fixed.

aidial_rag/documents.py
Allob: Fixed. Changed to `raise IndexIncompatibleError(f"Index format version is not supported: {doc_record.format_version}")`

aidial_rag/document_record_migration.py
Allob: Fixed.

aidial_rag/configuration_endpoint.py
adubovik: If the first person who saw this new parameter naming feels it's ambiguous, it likely proves that it wasn't a good idea. Let's leave it as is.

adubovik (approved):

Allob:
/deploy-review
>GitHub actions run: [17803518797](https://github.com/epam/ai-dial-ci/actions/runs/17803518797)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/687264)

Allob:
/deploy-review
>GitHub actions run: [17804865130](https://github.com/epam/ai-dial-ci/actions/runs/17804865130)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/687292)

Allob:
/deploy-review
>GitHub actions run: [17807773044](https://github.com/epam/ai-dial-ci/actions/runs/17807773044)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/687345)

Allob:
/deploy-review
>GitHub actions run: [17828456691](https://github.com/epam/ai-dial-ci/actions/runs/17828456691)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/687858)

Allob:
/deploy-review
>GitHub actions run: [17830443818](https://github.com/epam/ai-dial-ci/actions/runs/17830443818)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/687966)

Allob:
/deploy-review
>GitHub actions run: [17836275896](https://github.com/epam/ai-dial-ci/actions/runs/17836275896)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/688153)

Allob:
/deploy-review
>GitHub actions run: [17859352144](https://github.com/epam/ai-dial-ci/actions/runs/17859352144)
>- **Deployment failed**: [failed](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/688771)

Allob:
/deploy-review
>GitHub actions run: [17915277206](https://github.com/epam/ai-dial-ci/actions/runs/17915277206)
>Environment URL: [review-environment](https://chat-ai-dial-rag-pr-58.nightly-test.deltixhub.io) | [pipeline](https://gitlab.deltixhub.com/Clients/deltix/nightly-test/dial/review/-/pipelines/690037)

PR #118, opened 2026-02-26
fix: add support of scanned pdf with broken xref

### LINKED ISSUE #114: The PDF with broken xref table fails with The document is empty
### Name and Version

ai-dial-rag: 0.38.0

### What steps will reproduce the bug?

Add the PDF with broken xref table to the RAG, like [test_pdf_with_image_broken_xref.pdf](https://github.com/user-attachments/files/25397581/test_pdf_with_image_broken_xref.pdf)
The processing fails with a error: `The document is empty`

### What is the expected behavior?

RAG should be able to process broken document.

### What do you see instead?

Rag returns error: `The document is empty`

### Additional information

Some scanners produce PDF with incorrect xref table.

Some of the pdf libraries will fail to process such document, because it is incorrect. Some other libraries could work without xref table, by sequentially scanning the document and restore the offsets of the objects.
So, it is theoretically possible to support this type of documents in RAG.

Allob:
### Applicable issues

- fixes #114

### Description of changes

Add support of scanned pdf with broken xref table.
The pdfplumber library (and may others) is unable to read this pdf, because it is unable to find pages catalog using xref table.
But pdfplumber uses pypdfium2 to render page images. Pypdfium2 is able to read pdfs with broken xref by sequential scan.

This PR makes PdfPageImageExtractor to use pypdfium2 directly, without pdfplumber wrapper. This causes additional changes:
1. extract_number_of_pages is not async, to run pypdfium2 operations in single thread, since extract_number_of_pages  implementation using pypdfium2 is not thread-safe. This caused add_image_only_chunks and add_sources to be moved from get_document_chunks (non-async, within executor), to parse_document (async).

2. The extracted page images (and cached llm calls) for test_presentation_odp, test_presentation_pptm and test_presentation_pptm have changed.
Reason: converted pdf has pages defined with `/MediaBox[0 0 960.009448818898 540]`. We request image with width=800.
Before, pdfplumber would read the width as 960.009448818898 in python float type, calculated the scale and called pypdfium2. But pypdfium2 uses C float type, so for pypdfium2 the parsed width would be 960.0094604492188. Using the scale calculated by pdfplumber and [rounding up](https://github.com/pypdfium2-team/pypdfium2/blob/b82a5eefa6a67d7e3d7e243295ff29c4ef66304d/src/pypdfium2/_helpers/page.py#L447C21-L447C31), the resulting width of the image would be 801.
But now we use pypdfium2 directly and do not have this inconsistency - the resulting image width is 800.
The extracted images for the other tests have not changed.

### Checklist


Makefile
```diff
@@ -46,7 +46,7 @@ install_libreoffice:
 	sudo apt-get install --no-install-recommends -y libreoffice=$(LIBREOFFICE_UBUNTU_VERSION)
 
 test: install_nox $(if $(CI), install_libreoffice)
-	$(POETRY) run nox -s test $(ARGS)
+	$(POETRY) run nox -- -s test -- $(ARGS)
```
Allob, line 49: With poetry 2 we need `--` for the ARGS to be correctly passes to pytest and not nox.

tests/test_load_documents.py
```diff
@@ -48,6 +48,19 @@ async def test_load_pdf_with_image_and_no_text(local_server):
     assert chunks[0].metadata["page_number"] == 1
 
 
+@pytest.mark.asyncio
+async def test_load_pdf_with_broken_xref(local_server):
+    """Test loading a PDF with broken xref table
+
+    Some pdf libraries will unable to load this pdf, because they unable to find pages catalog using broken xref table and return empty document. But the PDF can be loaded successfully via sequential file scanning when the xref table is broken.
```
Allob, line 55: make lint made it a single line. It was multiline originally.

tests/test_image_extractor.py
```diff
@@ -65,3 +68,25 @@ async def test_attachment_image_invalid_page2():
...
+    images = await extract_pages(
+        mime_type="application/pdf",
+        file_bytes=pdf_bytes,
+        page_numbers=[1, 2],
+        scaled_size=800,
+    )
+
+    assert len(images) == 2
+    assert images[0].height == 450
+    assert images[0].width == 800
```
Allob, line 90: The width was 801 here with pdfplumber, because pdfplumber read the original page width as 960.009448818898, but pypdfium2 used C float inside and width was 960.009448818898. And pypdfium2 uses round up for rendering.

aidial_rag/document_loaders.py
```diff
@@ -249,10 +232,27 @@ async def parse_document(
...
             mime_type,
             attachment_link,
-            attachment_mime_type,
             parser_config,
         )
 
+        if are_image_pages_supported(mime_type):
+            # We will not have chunks from unstructured for the pages which does not contain text
+            # So we need to add them manually
+            chunks = await add_image_only_chunks(
```
Allob, line 241: `add_image_only_chunks()` uses `extract_number_of_pages()` which is async now. So it was moved from def `get_document_chunks()` to `async def parse_document()`

aidial_rag/image_processor/pdf_page_image_extractor.py
```diff
@@ -15,6 +14,50 @@
...
+        return len(pdf)
+
+
+def _render_page(
+    file_bytes: bytes,
+    page_number: int,
+    scaled_size: Optional[int] = None,
+) -> Image:
+    # Not thread safe because of pypdfium2
+    with closing(pdfium.PdfDocument(file_bytes)) as pdf:
```
Allob, line 42: The pdf is opened on every page render here. It is the same behavior as it was with pdfplumber.
There is a room for improvement here, but it could be a bit complicated because pypdfium2 is not thread-safe.
So, using a simple way here for now.

tests/test_image_extractor.py
```diff
@@ -65,3 +68,25 @@ async def test_attachment_image_invalid_page2():
                 file_bytes=pdf_bytes.read(),
                 page_numbers=[1, 1],
             )
+
+
+@pytest.mark.asyncio
+async def test_attachment_pptx():
```
adubovik: Misnomer `test_attachment_pptx`

aidial_rag/image_processor/pdf_page_image_extractor.py
```diff
@@ -48,19 +80,23 @@ async def extract_pages_gen(
...
+            if not (1 <= page_number <= total_pages):
+                raise RuntimeError(
+                    f"Invalid page number: {page_number}. Page number is ordinal number of the page. The document has {total_pages} pages."
                 )
-                logger.debug(f"Extracted page {page_number} as image")
-                yield image
+
+            logger.debug(f"Extracting page {page_number}...")
+
+            # Render in thread pool, because pypdfium2 is not thread safe
```
adubovik: `# Render in single thread pool, because pypdfium2 is not thread safe`

adubovik (dismissed):

adubovik (approved):

PR #128, opened 2026-03-04
chore: update unstructured to 0.21.5

Allob:
### Applicable issues

- fixes https://github.com/epam/ai-dial-rag/security/dependabot/70

### Description of changes

Nltk <= 3.9.2 have an issue, described in https://github.com/epam/ai-dial-rag/security/dependabot/70
As a result, unstructured started to use spaCy instead of nltk starting from unstructured=0.21.5.
This PR updates unstructured to 0.21.5 and provides the data necessary for unstructured  with spaCy to work.

We still need nltk for keywords search, because spaCy does not provide stemmer and has lemmatizer instead. Migration of keywords search to spaCy will break backwards compatibility with previous keywords search indexes.

Also, aidial-rag-eval was updated to 0.5.0, because new unstructured requires numpy 2, which is not supported by old aidial-rag-eval.

### Checklist


Dockerfile
```diff
@@ -55,9 +55,8 @@ RUN uvx "$POETRY" install --no-interaction --no-ansi --no-cache --only main --no
 
 FROM builder AS builder_download_nltk
 
-# nltk 3.9 actually uses punkt_tab and averaged_perceptron_tagger_eng
-# but we have to download punkt and averaged_perceptron_tagger as well, because unstructured will try to download it if missing
-RUN python -m nltk.downloader -d /usr/share/nltk_data stopwords punkt punkt_tab averaged_perceptron_tagger averaged_perceptron_tagger_eng
+# nltk stopwords and punkt_tab data is required for keywords search
+RUN python -m nltk.downloader -d /usr/share/nltk_data stopwords punkt_tab
```
Allob, line 59: For [keywords_search](https://github.com/epam/ai-dial-rag/blob/development/aidial_rag/keywords_search.py) we need just stopwords and punkt_tab.

tests/test_retrievers.py
```diff
@@ -60,7 +60,7 @@ async def test_retrievers(local_server):
         sys.stderr, buffer, mime_type, attachment_link, mime_type
     )
 
-    assert len(text_chunks) == 177
+    assert len(text_chunks) == 174
```
Allob, line 63: There are small changes in the resulting chunks:
For example, for two consecutive headers in side panel on page 1:
```
Geology

Alpine orogeny
```
The old version produces two separate chunks here, but in the new version this is a single chunk.

tests/cache/test_app/test_chat_completion_history/45e648a2ca0cca82ae47845371c0cf3b.response
```diff
@@ -0,0 +1,164 @@
...
+            },
+            {
+              "type": "text",
+              "text": "</doc>\n"
+            },
+            {
+              "type": "text",
+              "text": "<doc id='8' source='files/6iTkeGUs2CvUehhYLmMYXB/alps_wiki.html'>\nFauna\n\nThe Alps are a habitat for 30,000 species of wildlife, ranging from the tiniest snow fleas to brown bears, many of which have made adaptations to the harsh cold conditions and high altitudes to the point that some only survive in specific micro-climates either directly above or below the snow line.[3][64]\n\nThe largest mammal to live in the highest altitudes are the alpine ibex, which have been sighted as high as 3,000 m (9,843 ft). The ibex live in caves and descend to eat the succulent alpine grasses.[65] Classified as antelopes,[6] chamois are smaller than ibex and found throughout the Alps, living above the tree line and are common in the entire alpine range.[66] Areas of the eastern Alps are still home to brown bears. In Switzerland the canton of Bern was named for the bears but the last bear is recorded as having been killed in 1792 above Kleine Scheidegg by three hunters from Grindelwald.[67]\n"
```
Allob, line 114: There is a change in the LLM request because of the small changes in the BM25Retriever scores after nltk update.
There is a change in handling of a dash `\u2013`. With the old nltk dash did not split the tokens.
For example, `French\u2013Italian` and `192\u2013194` were single tokens, but now is tokenized as `["french", "\u2013", "italian"]` and `["192", "\u2013", "194"]`, respectively.

This changes the bm25 scores, and which chunk will be found among the chunks with lower relevance.

pyproject.toml
```diff
@@ -24,14 +24,13 @@ langchain-community = "^0.3.29"
 langchain-openai = "^0.3.11"
 aidial-sdk = {extras = ["telemetry"], version = "^0.21.0"}
 pypdf = "^6.6.0"
-beautifulsoup4 = "4.12.2"
```
Allob, line 27: We do not use beautifulsoup4 directly. Removed to allow unstructured to bring the version it needs.

Efreet408 (approved):

PR #186, opened 2026-07-23
ci: re-use venv

nepalevov:
CI speed-up:
- instruct **nox** to reuse the existing venv (created by **poetry**) instead of building a fresh one per session.

For local execution everything stays the same


nepalevov:
Too risky for now to mix llint/format/test tooling dependencies and project dependencies

